PR #7, opened 2019-10-15
Fix several compiler warnings

luator:
# Description
Fix several compiler warnings (e.g. about initialization order, signed vs unsigned, etc).

# How I Tested
Running some demo scripts on the Finger robot.

### DIFF
--- a/include/robot_interfaces/robot_data.hpp
+++ b/include/robot_interfaces/robot_data.hpp
@@ -55,7 +55,7 @@ public:
 
     RobotData(size_t history_length = 1000,
               bool use_shared_memory = false,
-              std::string shared_memory_address = "")
+              __attribute__((unused)) std::string shared_memory_address = "")
     {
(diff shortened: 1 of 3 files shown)


include/robot_interfaces/robot_data.hpp
```diff
@@ -55,7 +55,7 @@ class RobotData
 
     RobotData(size_t history_length = 1000,
               bool use_shared_memory = false,
-              std::string shared_memory_address = "")
+              __attribute__((unused)) std::string shared_memory_address = "")
```
mwuethri: i don't understand this, could you explain this?

luator: It suppresses the "unused variable" warning for this argument and makes it explicit in the code that we are aware that this is currently unused but still want to keep it for now.
I will add a comment in the code

PR #9, opened 2019-10-24
status as regular struct

vincentberenz:
These lines:

https://github.com/open-dynamic-robot-initiative/robot_interfaces/blob/master/include/robot_interfaces/robot_backend.hpp#L40

https://github.com/open-dynamic-robot-initiative/robot_interfaces/blob/master/include/robot_interfaces/robot_backend.hpp#L55

confuses me a lot, because on one hand RobotData is templated with Status, but there is no choice on what class Status should be. 

Also Status is indirectly templated as a member of RobotBackend<Action,Observation>, which makes it difficult to access in a "static" way.

For example, the RobotData I am using is templated, and declaration should looks like this:

```
  template<int NB_DOFS>
  class Data : public robot_interfaces::RobotData< PressureAction<NB_DOFS>,
                                                   State<NB_DOFS>,
                                                   robot_interfaces::RobotBackend<PressureAction<NB_DOFS>, State<NB_DOFS>>Status >

```

which does not compile because robot_interfaces::RobotBackend<PressureAction<NB_DOFS>, State<NB_DOFS>> is not a type.

By moving Struct as a simple member of the namespace, I have this:

```
  template<int NB_DOFS>
  class Data : public robot_interfaces::RobotData< PressureAction<NB_DOFS>,
                                                   State<NB_DOFS>,
                                                   robot_interfaces::Status >
  {};

```

which compiles fine.

So this pull requests changes from

`robot_interfaces::RobotBackend<Action,Observation>::Status
`
to simpler:
`
robot_interfaces::Status`

But I guess there must be some reason for this complicated design, and I certainly overlooked it ?

### DIFF
--- a/include/robot_interfaces/monitored_robot_driver.hpp
+++ b/include/robot_interfaces/monitored_robot_driver.hpp
@@ -168,9 +168,17 @@ private:
         // loop until shutdown and monitor action timing
         for (size_t t = 0; !is_shutdown_; t++)
         {
-            bool action_has_ended_on_time =
-                action_end_logger_.wait_for_timeindex(t,
-                                                      max_action_duration_s_);
+	  bool action_has_ended_on_time;
+	  if (max_action_duration_s_ >0)
+	    {
+	      action_has_ended_on_time = action_end_logger_.wait_for_timeindex(t,
+									       max_action_duration_s_);
+	    }
+	  else
+	    {
+	      action_has_ended_on_time = action_end_logger_.wait_for_timeindex(t);
+	    }
+	    
             if (!action_has_ended_on_time)
             {
                 std::cout
@@ -181,9 +189,15 @@ private:
                 return;
             }
 
-            bool action_has_started_on_time =
-                action_start_logger_.wait_for_timeindex(
-                    t + 1, max_inter_action_duration_s_);
+            bool action_has_started_on_time;
+	    if(max_inter_action_duration_s_>0)
+	      action_has_started_on_time = action_start_logger_.wait_for_timeindex(
+										   t + 1, max_inter_action_duration_s_);
+	    else
+	      {
+		action_has_started_on_time = action_start_logger_.wait_for_timeindex(
+										   t + 1);
+	      }
             if (!action_has_started_on_time)
             {
                 std::cout << "Action did not start on time, shutting down. Any "
--- a/include/robot_interfaces/finger_logger.hpp
+++ b/include/robot_interfaces/finger_logger.hpp
@@ -31,7 +31,6 @@ class FingerLogger
 public:
     typedef FingerTypes::Action Action;
     typedef FingerTypes::Observation Observation;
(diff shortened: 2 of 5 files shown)


vincentberenz:
@luator @mwuethri 
Ok, now that I moved to looking closer to the Frontend, I may be getting it. Creating instances of the backend with subclass of RobotData was maybe not the thing to do at all ...
Please ignore this while I am clarifying my mind :)

luator:
Moving `Status` out of the `RobotBackend` definitely makes sense (as you saw, there was already a TODO about this).  I'll have to double-check but I think there is no need for it to be templated (at least at the moment, may change in case we add more fields in the future).
I am a bit unsure at the moment, in which file it should be (ideally it should move out of `robot_backend.hpp` so that this class does not need to be included when only interested in the status struct). I'll take a look and update here.

luator:
After looking into the code, I think the answer about where and how `Status` should be defined mostly depends on how we are going to extend it in the future.

Right now `Status` is completely independent of the actual robot.  It should be defined outside the `RobotBackend` class.  I don't see an existing file where it really fits, so it probably should go to its own file.  As it is used by the backend, there is actually no option to use a different type here, so even the template parameter in `RobotData` for it could be removed, which would simplify the code a bit.

However, if we plan to extend the `Status` by (potentially robot-specific) fields in the future, we want to keep it where it is now.  The robot-specific Status would need to inherit from `RobotBackend::Status` because otherwise the backend would not work.  But in this case, I think it makes sense to keep the base Status inside the backend class to make clear that this something only relevant for the backend (the `RobotData` does not care if the status type is compatible with the backend or not).

In any case, I think the `RobotFrontend` should be extended by a `get_status` method so that the user can actually access that information.  For this, the `Status` type should be added to the template of the `RobotFrontend` class instead of typedefing it internally.

@mwuethri What do you think?

include/robot_interfaces/monitored_robot_driver.hpp
```diff
@@ -168,9 +168,17 @@ class MonitoredRobotDriver : public RobotDriver<Action, Observation>
...
+	  bool action_has_ended_on_time;
+	  if (max_action_duration_s_ >0)
+	    {
+	      action_has_ended_on_time = action_end_logger_.wait_for_timeindex(t,
+									       max_action_duration_s_);
+	    }
+	  else
+	    {
+	      action_has_ended_on_time = action_end_logger_.wait_for_timeindex(t);
+	    }
```
mwuethri: i would not handle this as a special case, and setting it to zero for allowing arbitrary action duration does not seem very logical to me. the idea is that you set it to infinity if you do not want it to be checked, then no special case needs to be added.

vincentberenz: @mwuethri setting the durations to infinity seems indeed cleaner.

vincentberenz:
closing this request and making a new one with cleaner fix

PR #12, opened 2019-10-28
Two joint driver

TheInfamousWayne:
added python bindings for running two joints with 1 motor-board

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,7 @@ add_subdirectory(demos)
 # python bindings #
 ###################
 
+# finger #
 set(py_finger_types_SRC_FILES
   srcpy/py_finger_types.cpp
 )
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -91,7 +92,7 @@ target_link_libraries(py_finger_types PRIVATE ${catkin_LIBRARIES})
 set_target_properties(py_finger_types PROPERTIES
     LIBRARY_OUTPUT_DIRECTORY ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_PYTHON_DESTINATION})
 install(TARGETS py_finger_types DESTINATION ${CATKIN_PACKAGE_PYTHON_DESTINATION})
-
+# one joint #
```
luator: Please keep an empty line between the blocks for different robots, that makes it a bit easier to read and handle.

luator:
@TheInfamousWayne Just to be sure, did you double-check if all files are included in this PR?

PR #13, opened 2019-10-29
Implementation of the generalisation of the logger and logging of the buffer data when cleanly terminated.

shrutij01:
TODO: change the way status is defined and move it to the n_joint_robot_types as a size one vector. But need to confirm this.


shrutij01:
Oh, and TODO: logging the status. Right now, the datatype conversions for the status are really messy so as and when the redefinition of status is confirmed (hopefully a size one vector), will add that line to the logger.

include/robot_interfaces/logger.hpp
```diff
@@ -0,0 +1,287 @@
...
+                        observation = (*logger_data_->observation)[j];
+                        status = (*logger_data_->status)[j];
+
+                        observation_data_ = observation.get_data();
+                        applied_action_data_ = applied_action.get_data();
+                        desired_action_data_ = desired_action.get_data();
+
+                        output_file_ << logger_data_->observation->timestamp_s(j) << " , " << j << " , " ;
+
+                        for(loop_iterator_ = 0; loop_iterator_ < observation_data_.size(); loop_iterator_++)
```
luator: As above, the loop variable can be defined locally.  Also the name is a bit misleading as an "iterator" is a concept in C++ that is different from using a loop variable.  I think in this case it is totally fine to just call it `i`.

Same for the loops below.

luator: Actually, with modern C++ it might be possible to do the following:
```
for (auto data: observation_data_)
{
    std::copy(data.begin(), data.end(), output_iterator_);
}
```
However, I don't know if this is supported by the time series.  If you want, you can give it a try, but the current approach of using a loop variable is also totally fine.

include/robot_interfaces/logger.hpp
```diff
@@ -0,0 +1,287 @@
...
+                        }
+
+                        for(loop_iterator_ = 0; loop_iterator_ < desired_action_data_.size(); loop_iterator_++)
+                        {
+                          std::copy(desired_action_data_[loop_iterator_].begin(), desired_action_data_[loop_iterator_].end(), output_iterator_);
+                        }
+
+                        output_file_ << std::endl;
+
+                        std::cout << j << std::endl;
```
luator: This was just for debugging, right? If not needed anymore, please remove it.

include/robot_interfaces/logger.hpp
```diff
@@ -0,0 +1,287 @@
...
+
+#ifdef VERBOSE
+                auto t1 = std::chrono::high_resolution_clock::now();
+#endif
+
+                for (long int j = index_; j < index_ + block_size_; j++)
+                {
+                    try
+                    {
+                        std::ostream_iterator<double> output_iterator_(output_file_, " , ");
```
luator: Did you find a solution for the trailing comma at the end of the line?

include/robot_interfaces/logger.hpp
```diff
@@ -0,0 +1,287 @@
...
+    {
+        thread_->create_realtime_thread(&FingerLogger::write, this);
+    }
+
+    void stop()
+    {
+      output_file_.open(output_file_name_, std::ios_base::app);
+      std::ostream_iterator<double> output_iterator_(output_file_, " , ");
+
+      for(int j = index_; j < logger_data_->observation->newest_timeindex(); j++)
```
luator: There is a lot of redundancy here with the `write` function.  To avoid this, you could move the redundant part into a separate function that is called by both.
I think the difference is mainly the loop condition.  You could use something like `min(index_ + block_size_, newest_timeindex())` which should work for both cases.

include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -15,6 +15,9 @@
 #include <robot_interfaces/robot_backend.hpp>
 #include <robot_interfaces/robot_data.hpp>
 #include <robot_interfaces/robot_frontend.hpp>
+#include <robot_interfaces/loggable.hpp>
+
+using namespace std;
```
luator: Please do not "use" any namespace in a header file.  The reason is that any file that includes this one will also be affected by this which can result in name collisions.

include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -27,12 +30,16 @@ namespace robot_interfaces
  *
  * @tparam N Number of joints
  */
+class Loggable;
```
luator: Is this forward declaration really needed (loggable.hpp is included above)?
Also, please do not add anything between a definition (of `NJointRobotTypes` in this case) and its corresponding doxygen comment block.

include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -27,12 +30,16 @@ namespace robot_interfaces
...
+
 template <size_t N>
 struct NJointRobotTypes
 {
     typedef Eigen::Matrix<double, N, 1> Vector;
+    //typedef std::vector<std::string> Field_Name;
+    //typedef std::vector<std::vector<double>> Field_Data;
```
luator: If not needed, please remove.

include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -43,6 +50,42 @@ struct NJointRobotTypes
...
+          Vector::Map(&position_kp_[0], position_kp.size()) = position_kp;
+
+          std::vector<double> position_kd_;
+          position_kd_.resize(position_kd.size());
+          Vector::Map(&position_kd_[0], position_kd.size()) = position_kd;
+
+          //then return them in a fixed size vector of vectors to avoid copying due to pushing back
+          //value of information!
+          std::vector<std::vector<double>> result;
+          result = {torque_, position_, position_kp_, position_kd_};
```
luator: This actually results in a copy (although the compiler may be smart enough to optimize it away).
To avoid it, you can first create the `result` vector with uninitialized entries of the required size and then write the values directly into it:
```
std::vector<std::vector<double>> result = {std::vector<double>(N), std::vector<double>(N), std::vector<double>(N), std::vector<double>(N)};
Vector::Map(&result[0][0], torque.size()) = torque;
```
(not tested but I think it should work like this).

include/robot_interfaces/logger.hpp
```diff
@@ -0,0 +1,287 @@
...
+    {
+      //using the newest index inofrmation in logger_data to extract information on the fields to be logged
+      applied_action = (*logger_data_->applied_action)[logger_data_->applied_action->newest_timeindex()];
+      desired_action = (*logger_data_->desired_action)[logger_data_->desired_action->newest_timeindex()];
+      observation = (*logger_data_->observation)[logger_data_->observation->newest_timeindex()];
+      status = (*logger_data_->status)[logger_data_->status->newest_timeindex()];
+
+      header.push_back("#");
+      header.push_back("[Timestamp]");
+      header.push_back("(Current Index)");
```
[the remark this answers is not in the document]
shrutij01: I just did that because it looked nice to me, seemed like differentiating the time information in the header looked cleaner. No specific reason. Will remove them if I should.

include/robot_interfaces/logger.hpp
```diff
@@ -0,0 +1,287 @@
...
+               !(logger_data_->desired_action->length() > 0))
+        {
+            real_time_tools::Timer::sleep_until_sec(0.1);
+        }
+
+        index_ = logger_data_->observation->newest_timeindex();
+
+        for (long int t = 0; !stop_was_called_; t++)
+        {
+            if (index_ + block_size_ ==
```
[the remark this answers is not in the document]
shrutij01: That's such a great point, thanks!

include/robot_interfaces/logger.hpp
shrutij01: Regarding the auto method, I currently implemented this hack for dealing with the trailing comma for which I have to use the loop variable anyway. So will implement this when I find a better solution for the comma. Thanks :)

include/robot_interfaces/logger.hpp
shrutij01: So, I have used an extremely hacky solution rn, just so that it works. I am basically using the same ostream_iterator only till the second last value. Then using another ostream_iterator with newline as the separator. Looking for a better solution.

include/robot_interfaces/logger.hpp
shrutij01: I didn't really get this and will ask you about it.

include/robot_interfaces/n_joint_robot_types.hpp
shrutij01: Great, thank you!

include/robot_interfaces/loggable.hpp
```diff
@@ -0,0 +1,34 @@
...
+* **/
+
+namespace robot_interfaces
+{
+class Loggable
+{
+public:
+
+  virtual std::vector<std::string> get_name() = 0;
+  virtual std::vector<std::vector<double>> get_data() = 0;
```
jviereck: Sort of random fly by comment: Should there also be timesteps corresponding to the recorded data?

luator: The timeseries already provides time information (the "time index" as well as a timestamp), so in general there should be no need to have it here.

include/robot_interfaces/logger.hpp
luator: I just found this [infix_iterator](https://stackoverflow.com/a/3497021/2095383) which is supposed to be a direct replacement for `std::ostream_iterator` without the trailing separator.
We could add this to the `mpi_cpp_tools` package.

include/robot_interfaces/loggable.hpp
jviereck: What's the information that should be represented in the `Loggable` object? For me, the contained data should also be the time index. This is not included in this struct yet.

I can see that the information is on the timeseries object. However, do you intend to forward this time information to the logger when using the Loggable interface?

shrutij01: The information that should be represented in the 'Loggable' object- the names of the fields in the 'Loggable' structure returned by the 'get_name()' method, and the fields themselves returned by 'get_data()'. This is being done so that we don't explicitly have to know what's inside the object. 

The time information is available to the logger by getting the timestamps associated with each element. So, there's no connection with Loggable for the time information.

I am also not sure I understood your question correctly, happy to discuss offline.

jviereck: Let's discuss offline. I think this is a more fundamental design discussion and better done on a whiteboard :)

shrutij01: Yes, great :)

PR #14, opened 2019-11-06
Add YAML conversion for Vector

luator:
# Description

Add a YAML conversion struct for `NJointRobotTypes::Vector` inside
`NJointRobotTypes` which can then easily be instantiated for the explicit
types.

Further explicitly define `OneJointTypes` and `TwoJointTypes` and add YAML conversions for the corresponding vectors.

# How I Tested
By using this to load robot configurations from YAML files (see https://github.com/open-dynamic-robot-initiative/blmc_robots/pull/24)

### DIFF
--- a/include/robot_interfaces/finger_types.hpp
+++ b/include/robot_interfaces/finger_types.hpp
@@ -20,3 +20,13 @@ struct FingerTypes : public NJointRobotTypes<3>
 };
 
 }  // namespace robot_interfaces
+
+
+namespace YAML
+{
+template <>
+struct convert<robot_interfaces::FingerTypes::Vector>
+    : robot_interfaces::FingerTypes::yaml_convert_vector
+{
+};
+}  // namespace YAML
--- a/include/robot_interfaces/n_joint_robot_types.hpp
+++ b/include/robot_interfaces/n_joint_robot_types.hpp
@@ -11,6 +11,7 @@
 #include <cmath>
 
 #include <Eigen/Eigen>
+#include <yaml-cpp/yaml.h>
 
 #include <robot_interfaces/robot_backend.hpp>
 #include <robot_interfaces/robot_data.hpp>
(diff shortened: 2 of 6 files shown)


include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -177,6 +178,53 @@ struct NJointRobotTypes
...
+     *
+     *     struct MyTypes : public NJointRobotTypes<42> {};
+     *
+     *     namespace YAML
+     *     {
+     *     template <>
+     *     struct convert<MyTypes::Vector> : MyTypes::yaml_convert_vector {};
+     *     }
+     */
+    struct yaml_convert_vector
```
mwuethri: i think this should probably not be here, since this is not specific to robot interfaces, and also the dependence on the template parameters can easily be avoided.
this should probably go to some generic package (e.g. mpi_cpp_tools)

include/robot_interfaces/finger_types.hpp
```diff
@@ -20,3 +20,13 @@ struct FingerTypes : public NJointRobotTypes<3>
 };
 
 }  // namespace robot_interfaces
+
+
+namespace YAML
+{
+template <>
+struct convert<robot_interfaces::FingerTypes::Vector>
+    : robot_interfaces::FingerTypes::yaml_convert_vector
```
mwuethri: i think i don't understand what's happening here, could you explain?

luator: This creates a convert struct for yaml-cpp as described at the bottom of [its tutorial page](https://github.com/jbeder/yaml-cpp/wiki/Tutorial).
All the implementation is done in `FingerTypes::yaml_convert_vector` but I have to bring it into the `YAML` namespace so I derive from it here to create a kind of alias.  There might be a more elegant solution, I am still lacking some understanding for the more complicated details of how templates work...

luator:
I did some more research and found that we already have a completely generic `Eigen::Matrix`-YAML conversion implementation here: https://github.com/machines-in-motion/yaml_cpp_catkin/blob/master/include/yaml_cpp_catkin/yaml_eigen.h

This already works for arbitrary sizes and makes this complete PR obsolete :)

PR #18, opened 2019-11-14
Generalisation of the Logger

shrutij01:
# A summary of the changes:
1. Stop method to save buffer data when the logger is cleanly killed.
2. New class loggable that defines methods to be implemented by any robot data structure that has to be externally accessed, for eg. for logging, so that explicit knowledge of the structure's fields is not required.
3. Implementation of these methods in NJointRobotTypes (Observation, and Action) and in the Backend (Status). Relevant changes in RobotLogger (Logger -> RobotLogger).
4. Infix_iterator to print the ", " delimiter between the elements of a line appended to the log file.

## Edit

4. is now outdated, as the infix_iterator doesn't iterate over vectors and currently returning a vector of vectors from get_data(). So, as discussed with @luator , instead using std :: ostream_iterator with (space) " " as a delimiter as there is no problem with a trailing space.


include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -161,11 +196,36 @@ struct NJointRobotTypes
...
+        std::vector<std::string> get_name() override
+        {
+            return {"Position", "Velocity", "Torque"};
+        }
+
+        std::vector<std::vector<double>> get_data() override
+        {
+            std::vector<double> position_;
+            position_.resize(position.size());
+            Vector::Map(&position_[0], torque.size()) = position;
```
luator: This uses `torques.size()` for the position.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+        int i, j;
+        std::string temp;
+
+        for (i = 0; i < field_name.size(); i++)
+        {
+          if(field_data[i].size() == 1)
+          {
+            temp = field_name[i];
+            header_.push_back(temp);
```
luator: I think it would be cleaner to pass `header_` as an argument to the function (using reference or pointer), then it does not need to be defined as a class variable but could be locally defined in the functions where it is used.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+    std::vector<std::string> header_;
+
+    int block_size_;
+    long int index_;
+
+    bool stop_was_called_;
+
+    std::ofstream output_file_;
+    std::string output_file_name_;
```
luator: I know this can be annoying but please add brief doxygen comments to all class variables and methods.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+            (*logger_data_->applied_action)[logger_data_->applied_action
+                                                ->newest_timeindex()];
+        desired_action =
+            (*logger_data_->desired_action)[logger_data_->desired_action
+                                                ->newest_timeindex()];
+        observation =
+            (*logger_data_
+                  ->observation)[logger_data_->observation->newest_timeindex()];
+        status =
+            (*logger_data_->status)[logger_data_->status->newest_timeindex()];
```
luator: The purpose of this code block is not clear to me.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+        return header_;
+
+    }
+
+    void append_name_to_header(std::vector<std::string> field_name,
+                               std::vector<std::vector<double>> field_data)
+    {
+
+        int i, j;
+        std::string temp;
```
luator: The scope of those can be reduced even more.
- i and j could be defined directly in the loop header and should actually of type `size_t`: `for (size_t i = 0; i < ...`
- temp can be defined where it is assigned (`std::string temp = field_name[i] ...`)

include/robot_interfaces/n_joint_robot_types.hpp
shrutij01: Omg, thanks. Fixed.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+            }
+          }
+        }
+
+    }
+
+    void append_header_to_file()
+    {
+
+      output_file_.open(output_file_name_, std::ios_base::app);
```
luator: Indentation does not follow our style guide.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+        output_file_.open(output_file_name_, std::ios_base::app);
+
+        std::vector<std::vector<double>> observation_data =
+            observation.get_data();
+        std::vector<std::vector<double>> applied_action_data =
+            applied_action.get_data();
+        std::vector<std::vector<double>> desired_action_data =
+            desired_action.get_data();
+        std::vector<std::vector<double>> status_data = status.get_data();
```
luator: Is this needed?  All variables are overwritten again inside the loop and since they are not used outside the loop, their definition could be moved inside.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+                std::cout << "Trying to access index older than the "
+                             "oldest! Skipping ahead."
+                          << std::endl;
+            }
+        }
+
+        output_file_.close();
+    }
+
+    void append_field_data_to_robot_data(std::vector<std::vector<double>> field_data)
```
luator: Shouldn't this be called `append_field_data_to_file`?

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+#ifdef VERBOSE
+                std::cout << "Index trying to access, oldest index in the "
+                             "buffer: "
+                          << j << ","
+                          << logger_data_->observation->oldest_timeindex()
+                          << std::endl;
+#endif
+
+// to print the time taken for one block of data to be logged.
+#ifdef VERBOSE
```
luator: The two `#ifdef` blocks can be combined.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,331 @@
...
+        logger_data_;
+
+    std::vector<std::string> header_;
+
+    int block_size_;
+    long int index_;
+
+    bool stop_was_called_;
+
+    std::ofstream output_file_;
```
luator: As the file is always closed at the end of each function where it is opened, I think it would be better to define this variable locally in each function instead of having it as a class member.

include/robot_interfaces/robot_logger.hpp
shrutij01: Yes, you are right.

include/robot_interfaces/robot_logger.hpp
shrutij01: It was an artefact of an older implementation which I forgot to check for after picking the commits. Removed.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -94,9 +96,19 @@ class RobotLogger
         thread_->join();
     }
 
+    /*
+    * get_header() is to get the title of the log file, describing all the information
+    * that will be logged in it.
+    */
+
```
luator: Please use the doxygen syntax for those comments (see [here](https://github.com/open-dynamic-robot-initiative/robot_interfaces/blob/7be66054198c70a47e303063417aeae619b76ecc/include/robot_interfaces/monitored_robot_driver.hpp#L88) for an example).
Also please omit the empty line between comment block and function.

include/robot_interfaces/robot_logger.hpp
```diff
@@ -233,16 +234,22 @@ class RobotLogger
...
+    void append_field_data_to_file(std::vector<std::vector<double>> field_data)
     {
 
       infix_ostream_iterator<double> double_iterator(output_file_, ", ");
 
       for (auto data : field_data)
       {
           std::copy(data.begin(),
-                    data.end(),
+                    data.end() - 1,
```
luator: What is the reason for the `- 1`?

include/robot_interfaces/robot_logger.hpp
```diff
@@ -0,0 +1,353 @@
...
+    void stop()
+    {
+
+        output_file_.open(output_file_name_, std::ios_base::app);
+
+        append_robot_data_to_file();
+
+        output_file_.close();
+
+    }
```
luator: I just noticed, this does not actually stop the logging.  The `stop_was_called_ = true;` and `thread_->join();` of the destructor should be moved to here.

include/robot_interfaces/robot_logger.hpp
shrutij01: Some fix I was trying for the comma problem, please ignore.

include/robot_interfaces/robot_logger.hpp
shrutij01: Yeah, I forgot to check again, had not cherry-picked the format commit. Sorry, fixed.

shrutij01:
**A couple of updates:**

1. I realised that the infix_iterator wasn't printing commas between two vectors from a vector of vectors (so when say, observation was passed to append_field_data_to_file(), there was no comma printed between the last value in the position vector, and the first value in the velocity vector). To fix this, we decided to do away with the ", " altogether and instead use " " to separate the values, both in the header and in each data line. In order to make parsing easier then, the names being returned by the get_name() method for any Loggable object *must* not contain spaces between words. 

2. In order to increase human-readability, I added an identifier in front of the names in the header so that it is easier to understand that the corresponding position/torque etc value is returned by the observation, or the applied action, or the desired action.

I am attaching a screenshot which shows how the log file looks like now.
Screenshot from 2019-11-15 10-55-21

include/robot_interfaces/robot_logger.hpp
shrutij01: Thanks, fixed.

include/robot_interfaces/robot_logger.hpp
shrutij01: I will look into this later.

include/robot_interfaces/robot_logger.hpp
shrutij01: Yes, thanks, implemented. :)

include/robot_interfaces/robot_logger.hpp
```diff
@@ -124,74 +115,78 @@ class RobotLogger
...
-    * according to the size of the field.
-    */
-
-    void append_name_to_header(std::string identifier, std::vector<std::string> field_name,
-                               std::vector<std::vector<double>> field_data)
+    /**
+     * @brief Fills in the name information of each field to be
+     * logged according to the size of the field.
+     *
+     * @param The structure the field corresponds to
```
luator: The doxygen [`@param`](http://www.doxygen.nl/manual/commands.html#cmdparam) tag expects the name of the parameter as first argument, i.e. it should be

    @param  identifier  The structure...

include/robot_interfaces/robot_logger.hpp
```diff
@@ -309,45 +303,38 @@ class RobotLogger
...
     {
-
-        output_file_.open(output_file_name_, std::ios_base::app);
-
+        stop_was_called_ = true;
         append_robot_data_to_file();
-
-        output_file_.close();
-
+        thread_->join();
```
luator: Please put the `thread_->joint()` above the `append_robot_data_to_file()` to prevent potential concurrency conflicts.

luator:
I just saw, there is a merge conflict in `py_finger_types.cpp`. I recently added a helper function for the bindings which is located in `pybind_helper.hpp`. Since the logger is now generic, you can add the bindings there and remove it completely from `py_finger_types.cpp`.

include/robot_interfaces/robot_logger.hpp
shrutij01: Sorry :|, fixed.

PR #20, opened 2019-11-19
Move Python package from `src/`  to `python/`

luator:
# Description
Move Python package from `src/` to `python/`.
This is to follow our style guide.

# How I Tested
Running some Python scripts without actual robot connected (just to verify the imports are working).


luator:
Resolved merge conflict

PR #22, opened 2019-11-29
Vberenz/status

vincentberenz:
#what
moved status to its own hpp file, directly in the robot interfaces namespace

#checked if ok
only checked success of compilation

### DIFF
--- a/include/robot_interfaces/status.hpp
+++ b/include/robot_interfaces/status.hpp
@@ -1,0 +1,27 @@
+#pragma once
+
+#include <string>
+#include <vector>
+#include <robot_interfaces/loggable.hpp>
+
+namespace robot_interfaces
+{
+
+  struct Status : public Loggable
+  {
+    
+    uint32_t action_repetitions = 0;
+    
+    std::vector<std::string> get_name() override
+    {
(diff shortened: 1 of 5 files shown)


include/robot_interfaces/status.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
luator: The file should probably have a license header like the other ones.

include/robot_interfaces/status.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <string>
+#include <vector>
+#include <robot_interfaces/loggable.hpp>
+
+namespace robot_interfaces
+{
+
+  struct Status : public Loggable
```
luator: Indentation is not following the style guide (4 spaces).

include/robot_interfaces/status.hpp
vincentberenz: indeed

include/robot_interfaces/status.hpp
vincentberenz: indeed

PR #23, opened 2019-11-29
Vberenz/demo

vincentberenz:
Minimal demo of front and backend, but on super dummy robot.

Would be great if you would confirm it corresponds to the intended usage of the API.

The demo works fine, but hang at the end. This corresponds to the behavior I described in a (still open) issue of the backend I created a while back. I will try to fix this.

### DIFF
--- a/demos/demo.cpp
+++ b/demos/demo.cpp
@@ -78,8 +78,8 @@ public:
   // at init dof are at min value
   void initialize()
   {
-    values_[0] = Driver::MIN;
-    values_[1] = Driver::MIN;
+    state_[0] = Driver::MIN;
+    state_[1] = Driver::MIN;
   }
 
   // just clip desired values
@@ -104,7 +104,7 @@ public:
 	// simulating the time if could take for a real
 	// robot to perform the action
 	usleep(1000);
-	values_[i] = applied.values[i];
+	state_[i] = applied.values[i];
       }
     return applied;
   }
@@ -112,8 +112,8 @@ public:
   Observation get_latest_observation()
   {
     Observation observation;
-    observation.values[0] = values_[0];
-    observation.values[1] = values_[1];
+    observation.values[0] = state_[0];
+    observation.values[1] = state_[1];
     return observation;
   }
     
@@ -121,7 +121,7 @@ public:
 
 private:
 
-  int values_[2];
+  int state_[2];
     
   const static int MAX = 1000;
   const static int MIN = 0;
@@ -168,7 +168,6 @@ int main()
       action.values[1]=500-value;
       // this action will be stored at index
       robot_interfaces::TimeIndex index = frontend.append_desired_action(action);
-      frontend.wait_until_timeindex(index+1);
       // getting the observation corresponding to the applied
       // action, i.e. at the same index
       observation = frontend.get_observation(index);
--- a/demos/demo.cpp
+++ b/demos/demo.cpp
@@ -1,0 +1,170 @@
+/**
+ * @file demo.cpp
+ * @author Vincent Berenz
+ * license License BSD-3-Clause
+ * @copyright Copyright (c) 2019, Max Planck Gesellschaft.
+ * 
+ * @brief Minimal demo of robot driver, backend and frontend
+ */
+
+#include "robot_interfaces/robot_driver.hpp"
+#include "robot_interfaces/robot_backend.hpp"
+#include "robot_interfaces/robot_frontend.hpp"
+
+#include <memory>
+
+/**
+ * \example demo.cpp
+ * This demo shows robot_interfaces of a 
+ * dummy "2dof" robot, in which a dof "position"
+ * is represented by an integer
+ */
+
+
+
+// Actions to be performed by robot, will be received by Driver
+// An action simply encapsulate two desired position value,
+// one for each dof
+class Action
+{
+  
+public:
+  
+  int values[2];
+
+  void print(bool backline)
+  {
+    std::cout << "action: "
+	      << values[0] << " "
+	      << values[1] << " ";
+    if(backline)
+      std::cout << "\n";
+  }
+  
+};
+
+// Read from the robot by Driver
+// An observation is the current position
+// for each dof
+class Observation
+{
+  
+public:
+  
+  int values[2];
+
+  void print(bool backline)
+  {
+    std::cout << "observation: "
+	      << values[0] << " "
+	      << values[1] << " ";
+    if(backline)
+      std::cout << "\n";
+  }
+  
+};
+
+
+// Send command to the robot and read observation from the robot
+// The dof positions simply becomes the ones set by the latest action,
+// capped between a min and a max value (0 and 1000)
+class Driver : public robot_interfaces::RobotDriver< Action,Observation>
+{
+  
+public:
+
+  Driver(){}
+
+  // at init dof are at min value
+  void initialize()
+  {
+    values_[0] = Driver::MIN;
+    values_[1] = Driver::MIN;
+  }
+
+  // just clip desired values
+  // between 0 and 1000
+  Action apply_action(const Action &desired_action)
+  {
+    Action applied;
+    for(unsigned int i=0;i<2;i++)
+      {
+	if(desired_action.values[i]>Driver::MAX)
+	  {
+	    applied.values[i]=Driver::MAX;
+	  }
+	else if(desired_action.values[i]<Driver::MIN)
+	  {
+	    applied.values[i]=Driver::MIN;
+	  }
+	else
+	  {
+	    applied.values[i]=desired_action.values[i];
+	  }
+	values_[i] = applied.values[i];
+      }
+    return applied;
+  }
+
+  Observation get_latest_observation()
+  {
+    Observation observation;
+    observation.values[0] = values_[0];
+    observation.values[1] = values_[1];
+    return observation;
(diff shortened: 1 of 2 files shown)


vincentberenz:
@mwuethri @luator 

also, as a side note, would be great to include such minimal demo everytime this is possible. Really helps first time users

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+class Action
+{
+  
+public:
+  
+  int values[2];
+
+  void print(bool backline)
+  {
+    std::cout << "action: "
```
mwuethri: i think the spacing here does not follow the conventions (4 spaces)

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+  {
+    values_[0] = Driver::MIN;
+    values_[1] = Driver::MIN;
+  }
+
+  // just clip desired values
+  // between 0 and 1000
+  Action apply_action(const Action &desired_action)
+  {
+    Action applied;
```
mwuethri: i think applied_action would be a bit more consistent naming

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+    observation.values[0] = values_[0];
+    observation.values[1] = values_[1];
+    return observation;
+  }
+    
+  void shutdown(){}
+
+private:
+
+  int values_[2];
```
mwuethri: maybe state_ would be a somewhat more descriptive name

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+	    applied.values[i]=Driver::MAX;
+	  }
+	else if(desired_action.values[i]<Driver::MIN)
+	  {
+	    applied.values[i]=Driver::MIN;
+	  }
+	else
+	  {
+	    applied.values[i]=desired_action.values[i];
+	  }
```
mwuethri: you could use the [clamp](https://github.com/open-dynamic-robot-initiative/mpi_cpp_tools/blob/master/include/mpi_cpp_tools/math.hpp#L17) function from mpi_cpp_tools to do all of the above in one line :)

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+int main()
+{
+
+  typedef robot_interfaces::RobotBackend<Action,Observation> Backend;
+  typedef Backend::Status Status;
+  typedef robot_interfaces::RobotData<Action,Observation,Status> Data;
+  typedef robot_interfaces::RobotFrontend<Action,Observation> Frontend;
+  
+  std::shared_ptr<Driver> driver_ptr(new Driver);
+  std::shared_ptr<Data> data_ptr(new Data);
```
mwuethri: not sure if new does the right thing... ususally they recomment std::make_shared

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+	else if(desired_action.values[i]<Driver::MIN)
+	  {
+	    applied.values[i]=Driver::MIN;
+	  }
+	else
+	  {
+	    applied.values[i]=desired_action.values[i];
+	  }
+	values_[i] = applied.values[i];
+      }
```
mwuethri: just to make the idea clear, you might want to wait 1ms here, since the idea is that it only returns once action has been completed

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+  typedef robot_interfaces::RobotBackend<Action,Observation> Backend;
+  typedef Backend::Status Status;
+  typedef robot_interfaces::RobotData<Action,Observation,Status> Data;
+  typedef robot_interfaces::RobotFrontend<Action,Observation> Frontend;
+  
+  std::shared_ptr<Driver> driver_ptr(new Driver);
+  std::shared_ptr<Data> data_ptr(new Data);
+
+  double max_action_duration_s = 1.0;
+  double max_inter_action_duration_s = 1.0;
```
mwuethri: max_action_duration_s should typically be slightly more than what you would expect to be the duration of your action

max_inter_action_duration_s should typically be small (unless you want a lot of tolerance)

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+  robot_interfaces::TimeIndex index = 0;
+  
+  for(uint value=200;value<=300;value++)
+    {
+      action.values[0]=value;
+      action.values[1]=500-value;
+      frontend.append_desired_action(action);
+      frontend.wait_until_timeindex(index+1);
+      observation = frontend.get_observation(index+1);
+      index++;
```
mwuethri: usually you would do 
TimeIndex t = frontend.append_desired_action(action);
observation = frontend.get_observation(t);

this way you don't need to handle the index yourself. the wait_until... is not required here since it will wait anyway when you get_observation.

to see the logic of the timestamps, see [here](https://github.com/open-dynamic-robot-initiative/robot_interfaces/blob/master/include/robot_interfaces/robot_data.hpp) the comments on top

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+  //robot_interfaces::TimeIndex index = frontend.get_current_timeindex();
+  robot_interfaces::TimeIndex index = 0;
+  
+  for(uint value=200;value<=300;value++)
+    {
+      action.values[0]=value;
+      action.values[1]=500-value;
+      frontend.append_desired_action(action);
+      frontend.wait_until_timeindex(index+1);
+      observation = frontend.get_observation(index+1);
```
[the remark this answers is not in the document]
luator: I think there is actually an issue with your implementation: The first action appended is assigned to time index 0. However, if you wait for `index+1` it will already expect the next action at that point, which has not been appended yet.  This should actually cause the robot to shut down (because the backend is not configured to repeat actions in this case).  However, this is currently not handled properly in the backend, I'll create an issue for this...

demos/demo.cpp
```diff
@@ -0,0 +1,171 @@
...
+    values_[1] = Driver::MIN;
+  }
+
+  // just clip desired values
+  // between 0 and 1000
+  Action apply_action(const Action &desired_action)
+  {
+    Action applied;
+    for(unsigned int i=0;i<2;i++)
+      {
```
luator: Indentation is inconsistent

demos/demo.cpp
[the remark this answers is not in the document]
mwuethri: wouldn't it shut down because inter_action_time would be exceeded?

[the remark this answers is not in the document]
luator: Hm, you are right. But only after the next action is appended, until then the backend would be blocked (the robot might proceed with the last action for that duration).  I think it would be much clearer and probably safer to add an explicit check for this in the backend.  Or is there a use-case where the current behaviour is needed?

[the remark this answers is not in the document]
luator: No, wait. I was confused.  It would not block, so everything is fine.  It is just not very clear when looking at the code.

demos/demo.cpp
luator: I was thinking about this as well. It would make it more readable but at the cost of adding a dependency and being less self-contained. So I am unsure which way is better.

demos/demo.cpp
vincentberenz: @mwuethri during the upcoming cleaning week, I plan to update all the code I developed to the convention.

demos/demo.cpp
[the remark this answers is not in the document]
luator: After offline discussion: While it does not block, it still opens a way for the user to mess up the backend timing by not providing actions fast enough, which is bad.  Will be addressed in #25.

demos/demo.cpp
vincentberenz: I do not like the idea of adding a dependency for such a trivial matter

demos/demo.cpp
vincentberenz: I have just read that it optimize times for memory allocation, i.e. in my code now there are 2 memory allocations in succession (one for driver, and one for the counter of the shared ptr), while when using std::make_shared, these 2 allocations occur in "one shot"

demos/demo.cpp
mwuethri: we already depend on this package, do you mean including the header?

vincentberenz:
@luator @mwuethri 
did all the requested changes, except for the clipping

demos/demo.cpp
```diff
@@ -152,17 +158,20 @@ int main()
...
+  // The other going from 300 to 200
   
   for(uint value=200;value<=300;value++)
     {
       action.values[0]=value;
       action.values[1]=500-value;
-      frontend.append_desired_action(action);
+      // this action will be stored at index
+      robot_interfaces::TimeIndex index = frontend.append_desired_action(action);
       frontend.wait_until_timeindex(index+1);
```
luator: This wait is not needed (`get_observation` will already wait internally).

demos/demo.cpp
vincentberenz: +1

demos/demo.cpp
vincentberenz: maybe, but I do not know this package enough to know what dependency for which reason. I just know that using an external package to perform a trivial clampling is not healthy. I also believe this a little too fine grained of a code review.

demos/demo.cpp
vincentberenz: removed !

PR #24, opened 2019-11-29
backend stops and exit by reapplying latest desired action

vincentberenz:
issue : backend hangs and its thread never join, because the loop hangs as the time series wait for new values that never comes

fix : destructor reapplies the latest desired action to unblock the loop

checking if it works : applied on the demo (see related pull request)

note : no unit tests for this, as failing unit test would mean unit test hanging forever


include/robot_interfaces/robot_backend.hpp
```diff
@@ -100,9 +107,10 @@ class RobotBackend
 private:
     MonitoredRobotDriver<Action, Observation> robot_driver_;
     std::shared_ptr<RobotData<Action, Observation, Status>> robot_data_;
-    bool destructor_was_called_;  // should be atomic
+    std::atomic<bool> should_stop_;
```
mwuethri: i find destructor_was_called_ more descriptive... but up to you guys :)

include/robot_interfaces/robot_backend.hpp
```diff
@@ -79,10 +80,16 @@ class RobotBackend
 
     virtual ~RobotBackend()
     {
-        destructor_was_called_ = true;
+        should_stop_ = true;
+	robot_data_->desired_action->append(desired_action_);
```
mwuethri: this may send an uninitialized double to the robot i think

include/robot_interfaces/robot_backend.hpp
```diff
@@ -128,12 +136,12 @@ class RobotBackend
 
         // wait until first desired_action was received
         // ----------------------------
-        while (!destructor_was_called_ &&
+        while (!should_stop_ &&
                !robot_data_->desired_action->wait_for_timeindex(0, 0.1))
         {
         }
```
mwuethri: i think actually the problem could be solved by just moving the block above into the for loop below, i.e. put

luator: Just moving the loop would not be enough but something similar should work.

@vincentberenz Let's talk about this on Monday.

luator: Not tested but I think something like the following should work:
```c++
while (!robot_data_->desired_action->wait_for_timeindex(t, 0.01)) {
    if (destructor_was_called_) {
        return;
    }
}

Action desired_action = (*robot_data_->desired_action)[t];
...
```
Note that I chose the timeout duration of 10ms rather arbitrarily, not sure what would be a good value.

mwuethri: yes, i think that should work. unlike what i said, i think we also need to leave the check before the loop. otherwise, i think we might have a similar problem in line 177 in the first iteration...

since this seems to be a quite common problem, i am starting to think that the best thing to do would be what @luator and me discussed: add a kill() function to Timeseries, which will make all the functions which are currently waiting for the next element throw an exception.

then we can simply kill all the timeseries in the destructor of the classes, and we immediately see if we did not handle the exception somewhere, and the hanging cannot happen anymore.

what do you think?

vincentberenz: @luator @mwuethri I implemented the solution suggested by Felix, and it worked, at least on the demo. I commited and push. I do not know the code enough to have an interesting opinion, but a kill function sounds useful. But in the meantime merging this would help, this is blocking for me.

include/robot_interfaces/robot_backend.hpp
```diff
@@ -188,6 +188,14 @@ class RobotBackend
 
             timers_[3].tic();
             // TODO: this may wait forever
+	    while (!destructor_was_called_ &&
```
luator: I think the `!destructor_was_called_` part should be removed. Otherwise there is the possible situation that the deconstructor is called just before this line is executed.  In that case, the while condition would be false and thus the internal check never executed.

mwuethri: oh yes, you are right!

mwuethri: nice catch!

vincentberenz: +1, corrected

include/robot_interfaces/robot_backend.hpp
```diff
@@ -100,9 +106,10 @@ class RobotBackend
 private:
     MonitoredRobotDriver<Action, Observation> robot_driver_;
     std::shared_ptr<RobotData<Action, Observation, Status>> robot_data_;
-    bool destructor_was_called_;  // should be atomic
+    std::atomic<bool> destructor_was_called_;
     uint32_t max_action_repetitions_;
-
+    Action desired_action_;
```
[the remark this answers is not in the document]
vincentberenz: that was a remnant of the previous implementation, my bad. Removed it.

include/robot_interfaces/robot_backend.hpp
```diff
@@ -83,6 +84,11 @@ class RobotBackend
         thread_->join();
     }
 
+    void stop()
+    {
+        destructor_was_called_ = true;
+    }
```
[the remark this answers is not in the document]
vincentberenz: also a remnant. removed it.

demos/demo.cpp
```diff
@@ -0,0 +1,180 @@
+/**
```
luator: Was this file added intentionally?

vincentberenz: I needed to have a least one instance that get destroyed, for testing. I removed it.

PR #26, opened 2019-12-10
Setting a precision value for the output data in the log file

shrutij01:
## What Changed

1. Earlier the logger was logging timestamps in the scientific notation, and now it has been set to log them with a precision of 27. I just set this randomly. 
2. Added a small note to the start() method basically highlighting that if the user specifies the same filename for logging data, then any further data whenever start() is called again will be appended to the same file. That works, else to specify another filename. 

Resolves open-dynamic-robot-initiative/blmc_robots/issues/38


luator:
This resolves open-dynamic-robot-initiative/blmc_robots/issues/38

PR #28, opened 2019-12-12
Remove Status from NJointRobotTypes

luator:
# Description
Apparently pybind11 has problems with creating multiple bindings of the same type.  Since `Status` is not templated anymore, it was exactly the same in all "types" modules.  This compiled without error but when importing two of the resulting modules (as happening in the `__init__.py`), it raised the following error: "ImportError: generic_type: type "Status" is already registered!".

## Changes done in the PR 
- To fix the issue mentioned above, move the binding of `Status` to a new module `py_generic`.
- To be consistent between Python and C++ remove the typedef from `NJointRobotTypes`.
- Add bindings for the constructor of `Status` and `Observation` so they can be instantiated in Python.

# How I Tested
- Verify `import robot_interfaces` does not raise an error.
- `rosrun blmc_robots demo_fake_finger.py`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,13 @@ target_link_libraries(demo ${catkin_LIBRARIES})
 # python bindings #
 ###################
 
+# generic #
+pybind11_add_module(py_generic srcpy/py_generic.cpp)
+target_link_libraries(py_generic PRIVATE ${catkin_LIBRARIES})
+set_target_properties(py_generic PROPERTIES
+    LIBRARY_OUTPUT_DIRECTORY ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_PYTHON_DESTINATION})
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -81,6 +81,13 @@ target_link_libraries(demo ${catkin_LIBRARIES})
 # python bindings #
 ###################
 
+# generic #
+pybind11_add_module(py_generic srcpy/py_generic.cpp)
```
vincentberenz: am I correct this will result in a python package called py_generic ? (as in : import py_generic). If so, are more specific name could be better.

luator: It will be `robot_interfaces.py_generic`, so not polluting the global namespace. There is also the `import *` in the `__init__.py`, so the user would never need to actually use that name (`from robot_interfaces import Status` would work).

luator:
Merging as the fix is needed (Manuel approved :))

PR #29, opened 2019-12-16
Add function process_desired_action

luator:
## Description

Add a function `process_desired_action()` which contains the processing
of the action (position control and safety-checks) which was originally
implemented in the `NJointBlmcRobotDriver` class in blmc_robots.

The purpose of moving it here in a separate function is to be able to
use it in other places as well and to be more testable.

For a cleaner interface wrap the function in a templated struct
`NJointRobotFunctions` similar to the `NJointRobotTypes`.

Also add unit tests for this function (implementing those already payed off:
It revealed a bug that custom gains set by the user were simply ignored.
This is already fixed in the function provided here.

## How I Tested
By running the new unit tests (`catkin run_tests robot_interfaces`).

### DIFF
--- a/include/robot_interfaces/n_joint_robot_functions.hpp
+++ b/include/robot_interfaces/n_joint_robot_functions.hpp
@@ -1,0 +1,131 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2019, Max Planck Gesellschaft
+// Copyright note valid unless otherwise stated in individual files.
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include "n_joint_robot_types.hpp"
+
+namespace robot_interfaces
+{
+/**
+ * @brief Collection of functions for a generic N-joint BLMC robot.
+ *
+ * Defines all the types needed to set up an interface to a generic N-joint BLMC
+ * robot that expects as Action a simple vector of N torque commands and
+ * provides N observations containing measured joint angle, velocity and torque.
+ *
+ * Defines some generic functions that can be used by all robots that are based
+ * on the `NJointRobotTypes`.
+ *
+ * @tparam N_JOINTS Number of joints the robot has.
+ */
+template <size_t N_JOINTS>
+struct NJointRobotFunctions
+{
+    using Types = NJointRobotTypes<N_JOINTS>;
+
+    /**
+     * @brief Process the desired action provided by the user.
+     *
+     * Takes the desired action from the user and does the following processing:
+     *
+     * ## 1. Run the position controller in case a target position is set.
+     *
+     *   If the target position is set to a value unequal to NaN for any joint,
+     *   a PD position controller is executed for this joint and the resulting
+     *   torque command is added to the torque command in the action.
+     *
+     *   If the P- and/or D-gains are set to a non-NaN value in the action, they
+     *   are used for the control.  NaN-values are replaced with the default
+     *   gains.
+     *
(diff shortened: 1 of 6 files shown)


tests/test_process_action.cpp
```diff
@@ -0,0 +1,345 @@
...
+                                          default_position_control_kp,
+                                          default_position_control_kd);
+
+    // joint 0 has positive velocity so expecting a reduced torque.
+    // joint 1 has negative velocity so expecting an increased torque.
+    // both should remain within the max. torque range.
+    ASSERT_LT(resulting_action.torque[0], 0.1);
+    ASSERT_GE(resulting_action.torque[0], -0.3);
+    ASSERT_GT(resulting_action.torque[1], 0.2);
+    ASSERT_LE(resulting_action.torque[1], 0.3);
```
mwuethri: also here it would be cleaner and more readable to use the variables. in fact i would probably put a for loop and check inside loop whether velocity for that joint is positive or negative and then do appropriate checks.

this way we are also robust to changes of values above. in fact it might make sense to run these tests which a bunch of different (random) values.

tests/test_process_action.cpp
```diff
@@ -0,0 +1,345 @@
...
```
mwuethri: same as above

tests/test_process_action.cpp
```diff
@@ -0,0 +1,345 @@
...
+                                          default_position_control_kp,
+                                          default_position_control_kd);
+
+    // Just testing here if the torque command goes in the right direction
+
+    ASSERT_LT(resulting_action.torque[0], 0.0);
+    ASSERT_GE(resulting_action.torque[0], -0.3);
+
+    ASSERT_GT(resulting_action.torque[1], 0.0);
+    ASSERT_LE(resulting_action.torque[1], 0.3);
```
mwuethri: same as above

tests/test_process_action.cpp
```diff
@@ -0,0 +1,345 @@
...
+    Types::Action resulting_action =
+        Functions::process_desired_action(action,
+                                          observation,
+                                          max_torque_Nm,
+                                          safety_kd,
+                                          default_position_control_kp,
+                                          default_position_control_kd);
+
+    ASSERT_EQ(0.1, resulting_action.torque[0]);
+    ASSERT_EQ(0.2, resulting_action.torque[1]);
```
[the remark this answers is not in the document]
luator: I replaced all plain numbers in asserts with the corresponding variables.

include/robot_interfaces/n_joint_robot_functions.hpp
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return Resulting action after applying all the processing.
+     */
+    static typename Types::Action process_desired_action(
+        const typename Types::Action &desired_action,
+        const typename Types::Observation &latest_observation,
+        const double max_torque_Nm,
+        const typename Types::Vector &safety_kd,
+        const typename Types::Vector &default_position_control_kp,
+        const typename Types::Vector &default_position_control_kd)
```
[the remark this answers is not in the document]
luator: I see that this is a bit dirty but I am not sure how to solve this. The only clean solution that I can see at the moment is to use different types for desired and applied action but this would add too much complexity to the interface in my opinion.

tests/test_process_action.cpp
luator: I can see some value in randomized tests but on the other hand, making the tests work with random input would make them very complex and thus difficult to write/read.  It would also increase the risk of having bugs in the tests themselves.
Another thing is, that in my opinion unit tests should be deterministic, otherwise this can be quite annoying for CI.

PR #32, opened 2020-01-09
Trigger error if next action is not provided in time

luator:
# Description
Trigger an error if the backend wants to apply the next action but none is available (and max. number of repetitions is exceeded).

Also add some documentation regarding the action repetition.

# How I Tested
Running demo_fake_finger.py.  When setting the number of repetitions to zero, the print of position happening after 300 iterations takes too long and causes the error as expected. When allowing repetitions (1 is already enough), the error does not occur.


luator:
@mwuethri Is there still some open issue here?

PR #33, opened 2020-01-10
Use time series from package "time_series"

luator:
# Description
Use the time series implementation from the new time_series package
instead of the one from real_time_tools.

Note that this should not change behaviour, it simply uses the implementation from the new package instead of the old one.

# How I Tested
By running demo_fake_finger.py from blmc_robots.

# Do not merge before
...treep config is updated: https://github.com/machines-in-motion/treep_machines_in_motion/pull/4 and corresponding merge request in git-amd.


luator:
@mwuethri Since vincent is on vacation, can you please take a look at this?

mwuethri:
whenever you have some time, could you do the same in blmc_drivers and blmc_robots, and then remove the ThreadsafeTimeseries from real_time_tools?

PR #34, opened 2020-01-13
Report timing errors to frontend

luator:
# Description
- Report errors from the timing watchdog in `MonitoredRobotDriver` through the `get_error()` method of the driver.
- Implement a `CheckpointTimer` to simplify the timing code in the backend loop.

The `CheckpointTimer` class should not be implemented where it is now.  If you approve the idea in general, I would move it to real_time_tools.

# Do not merge before
- [x] https://github.com/open-dynamic-robot-initiative/mpi_cpp_tools/pull/2
- [x] https://github.com/machines-in-motion/real_time_tools/pull/18

### DIFF
--- a/include/robot_interfaces/monitored_robot_driver.hpp
+++ b/include/robot_interfaces/monitored_robot_driver.hpp
@@ -10,9 +10,9 @@
 
 #include <iostream>
 
-#include <mpi_cpp_tools/multithreading.hpp>
 #include <real_time_tools/process_manager.hpp>
 #include <real_time_tools/thread.hpp>
+#include <real_time_tools/threadsafe/threadsafe_object.hpp>
 #include <real_time_tools/timer.hpp>
 #include <time_series/time_series.hpp>
 
@@ -163,7 +163,7 @@ private:
 
     std::shared_ptr<real_time_tools::RealTimeThread> thread_;
 
-    MutexedType<std::string> error_message_;
+    real_time_tools::SingletypeThreadsafeObject<std::string, 1> error_message_;
 
     /**
      * @brief Monitor the timing of action execution.
--- a/include/robot_interfaces/monitored_robot_driver.hpp
+++ b/include/robot_interfaces/monitored_robot_driver.hpp
@@ -10,6 +10,7 @@
 
 #include <iostream>
 
+#include <mpi_cpp_tools/multithreading.hpp>
 #include <real_time_tools/process_manager.hpp>
 #include <real_time_tools/thread.hpp>
 #include <real_time_tools/timer.hpp>
@@ -124,7 +125,7 @@ public:
         const std::string driver_error = robot_driver_->get_error();
         if (driver_error.empty())
         {
-            return error_message_;
+            return error_message_.get();
         }
         else
         {
@@ -162,7 +163,7 @@ private:
 
     std::shared_ptr<real_time_tools::RealTimeThread> thread_;
 
-    std::string error_message_ = "";
+    MutexedType<std::string> error_message_;
 
     /**
      * @brief Monitor the timing of action execution.
@@ -188,7 +189,7 @@ private:
                                                       max_action_duration_s_);
             if (!action_has_ended_on_time)
             {
-                error_message_ = "Action did not end on time, shutting down.";
+                error_message_.set("Action did not end on time, shutting down.");
                 shutdown();
                 return;
             }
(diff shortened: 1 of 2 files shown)


include/robot_interfaces/robot_backend.hpp
```diff
@@ -26,12 +26,113 @@
...
+
+    /**
+     * @brief Set checkpoint for time measurement.
+     *
+     * Measures time from the last call of start() or checkpoint() until this
+     * call.  The given name is used when printing the results.
+     *
+     * @param checkpoint_name Name of the checkpoint (used for printing results)
+     */
+    void checkpoint(const std::string &checkpoint_name)
```
mwuethri: isn't there a problem here if i don't call the checkpoints always in the same order? i.e. if all the names have been set, and the next time the user calls this function in a different order, it should throw an error i guess.

so probably a check of checkpoint_name_[current_checkpoint] == checkpoint_name would be necessary.

include/robot_interfaces/robot_backend.hpp
```diff
@@ -26,12 +26,113 @@
...
+            {
+                std::cout << "----- " << checkpoint_names_[i] << std::endl;
+                timers_[i].print_statistics();
+            }
+        }
+    }
+
+private:
+    std::array<real_time_tools::Timer, NUM_CHECKPOINTS + 1> timers_;
+    std::array<std::string, NUM_CHECKPOINTS + 1> checkpoint_names_;
```
mwuethri: i think checkpoint_names_[0] is never used. how about setting the name to "total" ? then you also don't need to make a special case in print_statistics()

include/robot_interfaces/robot_backend.hpp
```diff
@@ -26,12 +26,113 @@
...
+ *     timer.checkpoint("logging");
+ * }
+ * ~~~
+ *
+ * @tparam NUM_CHECKPOINTS Number of checkpoints.
+ * @tparam ENABLED Set to false, to disable timer.  Method calls will have no
+ * effect (and should hopefully be optimized away by the compiler).
+ */
+template <size_t NUM_CHECKPOINTS, bool ENABLED = true>
+class CheckpointTimer
```
mwuethri: i think this class is very useful! maybe it would make sense to integrate it into the real_time_tools::Timer class?

luator: Not sure if integrating it into the `Timer` class makes sense (it uses this class internally) but I am going to move it to the `real_time_tools` package.

include/robot_interfaces/robot_backend.hpp
luator: :+1:

include/robot_interfaces/robot_backend.hpp
luator: Do you think this is a realistic scenario? I would like to keep the timing code as compact as possible to not delay the loop too much, as this is affecting the "Total" measurement.

luator: Hm, something that might actually happen is a checkpoint that is accidentally put inside an `if` and thus is not executed in every iteration...

include/robot_interfaces/monitored_robot_driver.hpp
```diff
@@ -147,13 +155,15 @@ class MonitoredRobotDriver : public RobotDriver<Action, Observation>
...
     //! \brief Whether shutdown was initiated.
-    bool is_shutdown_;  // TODO: should be atomic
+    std::atomic<bool> is_shutdown_;
 
     real_time_tools::ThreadsafeTimeseries<bool> action_start_logger_;
     real_time_tools::ThreadsafeTimeseries<bool> action_end_logger_;
 
     std::shared_ptr<real_time_tools::RealTimeThread> thread_;
 
+    std::string error_message_ = "";
```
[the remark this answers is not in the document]
luator: Right, I will add it.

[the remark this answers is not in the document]
luator: done

luator:
@mwuethri I addressed all comments.  Only need to wait until corresponding PRs on other repos are merged now.

mwuethri:
thanks! feel free to merge whenever you're ready, i already approved.

luator:
@mwuethri I switched to the `real_time_tools::SingletypeThreadsafeObject` for the mutex of `error_message_`.  As that class seems to be quite complex, could you quickly check if I use it correctly?

mwuethri:
looks good, but i think this class hasn't been used for a while, so I am not sure if we should trust that code 100%...

luator:
I looked at the code of `SingletypeThreadsafeObject`. The part I am using (only `get` and `set`) seems to implement more or less the same as the class I used before.  So I think it should be alright.

PR #35, opened 2020-01-14
added Robot, which encaspulate frontend and backend

vincentberenz:
# What it is

The class Robot is a subclass of Frontend, which encapsulates a Backend.
This allows to construct a Frontend and a Backend in a single instance. This instance provides the same API as Frontend.

edit : In a supplementary commit, Robot takes the class RobotDriver as template, and construct the related instance. So, the frontend, the backend, the data and the driver instances are all instantiated via a single call.

# How it was tested

demo.cpp was extended to also use Robot

### DIFF
--- a/demos/demo.cpp
+++ b/demos/demo.cpp
@@ -145,7 +145,6 @@ int main()
 				      Observation,
 				      robot_interfaces::Status> Data;
   typedef robot_interfaces::RobotFrontend<Action,Observation> Frontend;
-  typedef robot_interfaces::Robot<Action,Observation> Robot;
   
   // max time allowed for the robot to apply an action.
   double max_action_duration_s = 0.02;
@@ -197,11 +196,12 @@ int main()
   {
 
     std::cout << "\n -- * -- Robot -- * --\n" << std::endl;
+
+    typedef robot_interfaces::Robot<Action,Observation,Driver> Robot;
     
-    std::shared_ptr<Driver> driver_ptr = std::make_shared<Driver>();
-    Robot robot(driver_ptr,
-		max_action_duration_s,
+    Robot robot(max_action_duration_s,
 		max_inter_action_duration_s);
+    
     robot.initialize();
 
     Action action;
--- a/demos/demo.cpp
+++ b/demos/demo.cpp
@@ -75,13 +75,14 @@ class Driver : public robot_interfaces::RobotDriver< Action,Observation>
   
 public:
 
-  Driver(){}
+  Driver(int min, int max)
+    : min_(min),max_(max){}
 
   // at init dof are at min value
   void initialize()
   {
-    state_[0] = Driver::MIN;
-    state_[1] = Driver::MIN;
+    state_[0] = min_;
+    state_[1] = min_;
   }
 
   // just clip desired values
@@ -91,13 +92,13 @@ public:
     Action applied;
     for(unsigned int i=0;i<2;i++)
       {
-	if(action_to_apply.values[i]>Driver::MAX)
+	if(action_to_apply.values[i]>max_)
 	  {
-	    applied.values[i]=Driver::MAX;
+	    applied.values[i]=max_;
 	  }
-	else if(action_to_apply.values[i]<Driver::MIN)
+	else if(action_to_apply.values[i]<min_)
 	  {
-	    applied.values[i]=Driver::MIN;
+	    applied.values[i]=min_;
 	  }
 	else
 	  {
@@ -129,9 +130,8 @@ public:
 private:
 
   int state_[2];
-    
-  const static int MAX = 1000;
-  const static int MIN = 0;
+  int min_;
+  int max_;
 
 };
 
@@ -158,7 +158,7 @@ int main()
 
     std::cout << "\n -- * -- Frontend and Backend -- * --\n" << std::endl;
     
-    std::shared_ptr<Driver> driver_ptr = std::make_shared<Driver>();
+    std::shared_ptr<Driver> driver_ptr = std::make_shared<Driver>(0,1000);
     std::shared_ptr<Data> data_ptr = std::make_shared<Data>();
 
     Backend backend(driver_ptr,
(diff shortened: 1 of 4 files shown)


include/robot_interfaces/robot.hpp
```diff
@@ -0,0 +1,86 @@
...
+	DataPtr data_ptr = std::make_shared<Data>())
+    : RobotFrontend<Action,Observation>(data_ptr),
+    data_ptr_(data_ptr),
+    driver_ptr_(std::make_shared<Driver>(args ...)),
+    backend_(driver_ptr_,
+	     data_ptr_,
+	     max_action_duration_s,
+	     max_inter_action_duration_s)
+  {
+    // compile time checking template Driver inherate from RobotDriver
```
luator: Typo "inherate"

include/robot_interfaces/robot.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * @brief RobotFrontend that construct and encapsulates
+ * its related RobotBackend. It also construct and starts
+ * the robot driver.
+ */
+template<typename Action,
+	 typename Observation,
+	 typename Driver,
+	 typename ... Args>
```
luator: It might be good extending the demo a bit to make use of this feature.

vincentberenz: @luator 
Not sure what you mean, I also updated the demo to perform the two version (frontend/backend, and robot)

demos/demo.cpp
```diff
@@ -145,42 +146,84 @@ int main()
...
+	action.print(false);
+	observation.print(true);
+      }
+  }
+
+  // demo representing usage of frontend and backend
+  // encapsulated in the same instance
+  {
+
+    std::cout << "\n -- * -- Robot -- * --\n" << std::endl;
```
vincentberenz: robot verson of the demo here

include/robot_interfaces/robot.hpp
luator: I mean the arguments for the driver (the `... Args` part).  In the demo, the driver does not take any arguments.  I am not familiar with this syntax, I assume it would work like the following?

    robot_interfaces::Robot<Action, Observation, Driver, int, double> robot(max_duration, max_inter, 42, 3.14);

vincentberenz: Not intuitively, it would work as:

robot_interfaces::Robot<Action, Observation, Driver> robot(max_duration, max_inter, 42, 3.14);

+1 on updating the demo, will do

luator:
What is the status here?

vincentberenz:
I had to move to something else for a while, but this is still in progress. There are some bugs in the latest commit I found out and that I am correcting

demos/demo.cpp
```diff
@@ -74,13 +75,14 @@ class Driver : public robot_interfaces::RobotDriver< Action,Observation>
   
 public:
 
-  Driver(){}
+  Driver(int min, int max)
```
vincentberenz: This shows an example of a Driver taking arguments

demos/demo.cpp
```diff
@@ -145,42 +146,88 @@ int main()
...
+
+    std::cout << "\n -- * -- Robot -- * --\n" << std::endl;
+
+    typedef robot_interfaces::Robot<Action,Observation,Driver> Robot;
+
+    int min = 0;
+    int max = 100;
+    Robot robot(max_action_duration_s,
+		max_inter_action_duration_s,
+		min,
```
vincentberenz: Arguments for the driver are simply passed to the constructor of Robot

include/robot_interfaces/monitored_robot_driver.hpp
```diff
@@ -172,32 +175,45 @@ class MonitoredRobotDriver : public RobotDriver<Action, Observation>
...
-            {
-                std::cout << "Action did not start on time, shutting down. Any "
-                             "further actions will be ignored."
-                          << std::endl;
-                shutdown();
-                return;
-            }
-        }
+	  {
+	    if(! std::isinf(max_action_duration_s_))
```
vincentberenz: This update is to make the class robust to infinite durations (without these changes it would hang indefinitely on hand). Note that this change is not related to the initial purpose of this pull request. Creating a new pull request sounds too much overhead to me, but will never do it if requested

include/robot_interfaces/robot.hpp
vincentberenz: done

vincentberenz:
@luator 

"LGTM but please reformat the code before merging."

I ran the reformating (as implemented to mpi-cmake-modules), but this results in changes in the whole repo. What wuuld be the best way to proceed ?

include/robot_interfaces/monitored_robot_driver.hpp
```diff
@@ -172,32 +172,45 @@ class MonitoredRobotDriver : public RobotDriver<Action, Observation>
...
+		      << "Action did not end on time, shutting down. Any further "
+		      "actions will be ignored."
+		      << std::endl;
+		    shutdown();
+		    return;
+		  }
+	      }
+	      else
+		{	    
+		  real_time_tools::Timer::sleep_ms(1);
```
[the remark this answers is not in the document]
vincentberenz: oops, you are right. will do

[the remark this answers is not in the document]
luator: Did you have time to look into this already?  I am actually running into the same issue right now (monitored driver hanging on destruction with infinite timeouts).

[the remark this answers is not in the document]
luator: I added a check in the monitored driver and simply don't start the thread if both timeouts are infinite ([see here](https://github.com/open-dynamic-robot-initiative/robot_interfaces/blob/d15cc3eb32396b80e801a1b3e611e290145c653f/include/robot_interfaces/monitored_robot_driver.hpp#L76)).  The corresponding PR will hopefully soon be merged, so you could simply wait for this.

However, if you already have a nicer solution, I'm happy to discard my fix and adopt yours :)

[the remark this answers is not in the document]
vincentberenz: done, I also applied clang_format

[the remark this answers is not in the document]
luator: Oh, I just saw, that you already pushed something regarding this.

vincentberenz:
@luator merged !

PR #36, opened 2020-01-21
Enable usage of multi-process time series in RobotData

luator:
# Summary
- Split RobotData into a single-process and multi-process version.
- Add a C++ demo for multiprocessing.

# Detailed
The `RobotData` class remains as base class, holding the pointers to the
time series, but does not initialize them.  Initialization is done in
the derived classes `SingleProcessRobotData` (using the single process
time series) and `MultiProcessRobotData` (using the multi process time
series).

This split is needed because the originally intended design with a
if-check does not compile with template types that are not serializable,
even when only single process is used.

While doing this, also remove the `Status` argument from the templates
(the type for this is fixed anyway).

Add a multiprocess demo. It is based on demo.cpp but split into separate executables for
backend and frontend, using the `MultiProcessRobotData` for communication.

Also do some cleanup in the CMakeLists.txt.

# How I Tested
By running the demos of this package as well as the `demo_fake_finger.py` of blmc_robots (adapt the latter to work with single- and multi-process versions of robot data).

# Do not merge before
- [x] https://github.com/open-dynamic-robot-initiative/blmc_robots/pull/50 is ready.
- [ ] ~corresponding merge request on blmc_robots_ros~

### DIFF
--- a/include/robot_interfaces/cereal_eigen.hpp
+++ b/include/robot_interfaces/cereal_eigen.hpp
@@ -7,7 +7,8 @@
  * @license CC BY-SA 4.0
  * @todo Move this to some "serialization tools" package.
  *
- * Taken from https://stackoverflow.com/a/51944389/2095383.
+ * Taken from https://stackoverflow.com/a/51944389/2095383 with minor
+ * modifications.
  */
 #include <Eigen/Eigen>
 #include <type_traits>
@@ -19,12 +20,22 @@ inline typename std::enable_if<
     traits::is_output_serializable<BinaryData<typename Derived::Scalar>,
                                    Archive>::value,
     void>::type
-save(Archive& ar, Eigen::PlainObjectBase<Derived> const& m)
+save(Archive& archive, const Eigen::PlainObjectBase<Derived>& object)
 {
     typedef Eigen::PlainObjectBase<Derived> ArrT;
-    if (ArrT::RowsAtCompileTime == Eigen::Dynamic) ar(m.rows());
-    if (ArrT::ColsAtCompileTime == Eigen::Dynamic) ar(m.cols());
-    ar(binary_data(m.data(), m.size() * sizeof(typename Derived::Scalar)));
+
+    // only add dimensions to the serialized data when they are dynamic
+    if (ArrT::RowsAtCompileTime == Eigen::Dynamic)
+    {
+        archive(object.rows());
+    }
+    if (ArrT::ColsAtCompileTime == Eigen::Dynamic)
+    {
+        archive(object.cols());
+    }
+
+    archive(binary_data(object.data(),
+                        object.size() * sizeof(typename Derived::Scalar)));
 }
 
 template <class Archive, class Derived>
@@ -32,15 +43,24 @@ inline typename std::enable_if<
     traits::is_input_serializable<BinaryData<typename Derived::Scalar>,
                                   Archive>::value,
     void>::type
-load(Archive& ar, Eigen::PlainObjectBase<Derived>& m)
+load(Archive& archive, Eigen::PlainObjectBase<Derived>& object)
 {
     typedef Eigen::PlainObjectBase<Derived> ArrT;
+
     Eigen::Index rows = ArrT::RowsAtCompileTime, cols = ArrT::ColsAtCompileTime;
-    if (rows == Eigen::Dynamic) ar(rows);
-    if (cols == Eigen::Dynamic) ar(cols);
-    m.resize(rows, cols);
-    ar(binary_data(m.data(),
-                   static_cast<std::size_t>(rows * cols *
-                                            sizeof(typename Derived::Scalar))));
+    // information about dimensions are only serialized for dynamic-size types
+    if (rows == Eigen::Dynamic)
+    {
+        archive(rows);
+    }
+    if (cols == Eigen::Dynamic)
+    {
+        archive(cols);
+    }
+
+    object.resize(rows, cols);
+    archive(binary_data(object.data(),
+                        static_cast<std::size_t>(
+                            rows * cols * sizeof(typename Derived::Scalar))));
 }
 }  // namespace cereal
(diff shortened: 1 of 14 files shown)


include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -19,6 +19,33 @@
 
 #include <robot_interfaces/robot_logger.hpp>
 
+// https://stackoverflow.com/a/51944389/2095383
+// Authors: Azoth, eudoxos
+// Date: 2020-01-15
+// License: CC BY-SA 4.0
+namespace cereal
```
luator: I just noticed that these serialization functions are still here. They should be moved to a more central place as it might be useful in other places.  I'll do this and update here.

include/robot_interfaces/cereal_eigen.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <type_traits>
+
+namespace cereal
+{
+template <class Archive, class Derived>
+inline typename std::enable_if<
+    traits::is_output_serializable<BinaryData<typename Derived::Scalar>,
+                                   Archive>::value,
+    void>::type
+save(Archive& ar, Eigen::PlainObjectBase<Derived> const& m)
```
mwuethri: variable names could be a bit more descriptive

include/robot_interfaces/cereal_eigen.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+    typedef Eigen::PlainObjectBase<Derived> ArrT;
+    if (ArrT::RowsAtCompileTime == Eigen::Dynamic) ar(m.rows());
```
mwuethri: i don't understand what this does

include/robot_interfaces/cereal_eigen.hpp
mwuethri: why is the const here after the type? is this the same as const Eigen::PlainObjectBase<Derived> &?

include/robot_interfaces/cereal_eigen.hpp
```diff
@@ -0,0 +1,46 @@
...
+    if (ArrT::ColsAtCompileTime == Eigen::Dynamic) ar(m.cols());
+    ar(binary_data(m.data(), m.size() * sizeof(typename Derived::Scalar)));
+}
+
+template <class Archive, class Derived>
+inline typename std::enable_if<
+    traits::is_input_serializable<BinaryData<typename Derived::Scalar>,
+                                  Archive>::value,
+    void>::type
+load(Archive& ar, Eigen::PlainObjectBase<Derived>& m)
```
mwuethri: shouldn't the first input here be const?

include/robot_interfaces/cereal_eigen.hpp
```diff
@@ -0,0 +1,46 @@
...
+ *        with cereal.
+ * @authors Azoth, eudoxos
+ * @date 2020-01-15
+ * @license CC BY-SA 4.0
+ * @todo Move this to some "serialization tools" package.
+ *
+ * Taken from https://stackoverflow.com/a/51944389/2095383.
+ */
+#include <Eigen/Eigen>
+#include <type_traits>
```
mwuethri: these functions are a little but obscure to me. one important question: is there any problem with memory allocation by using these functions? in particular, by using this stuff inside the timeseries, is any memory being allocated? that would be an issue.

include/robot_interfaces/n_joint_robot_types.hpp
```diff
@@ -48,6 +49,12 @@ struct NJointRobotTypes
         //! D-gain for position controller.  If NaN, default is used.
         Vector position_kd;
 
+        template <class Archive>
+        void serialize(Archive& archive)
+        {
+            archive(torque, position, position_kp, position_kd);
```
mwuethri: similar question here: are we sure this does not allocate memory? @vincentberenz @MaximilienNaveau, i guess you guys had to deal with these kind of questions.

include/robot_interfaces/cereal_eigen.hpp
luator: Yes, [apparently it is the same](https://stackoverflow.com/q/3694630/2095383) and it's just a matter of style.

include/robot_interfaces/cereal_eigen.hpp
luator: It only adds the size of the matrix/array to the serialized data if it is a dynamic-size type and thus this information is needed to deserialise.  I.e. for something like a Vector3d the size is omitted as it is anyway known.

include/robot_interfaces/cereal_eigen.hpp
luator: Not sure if it would work, cereal gives this signature for the load function in its documentation.

include/robot_interfaces/cereal_eigen.hpp
luator: done

include/robot_interfaces/cereal_eigen.hpp
luator: This is indeed a valid concern.  I searched for a while but did not find any information on this (I also looked into the code but did not understand enough of it).  Maybe @vincentberenz knows more about this?

include/robot_interfaces/n_joint_robot_types.hpp
luator: @mwuethri Is it okay if we merge this for now?  In case there is really some memory allocation issue in the serialization, this would anyway be a more fundamental issue that would not be fixed in this PR.

include/robot_interfaces/cereal_eigen.hpp
vincentberenz: @luator @mwuethri What function are you mentionning ? The code snippet above shows some include

include/robot_interfaces/n_joint_robot_types.hpp
vincentberenz: My understanding of cereal : it is highly templated exactly to avoid runtime memory allocation

include/robot_interfaces/cereal_eigen.hpp
luator: @vincentberenz Sorry, I missed your question earlier.  At least I was referring to cereal in general which you basically answered below.

PR #39, opened 2020-03-09
Storing images from a webcam using OpenCV

shrutij01:
## Description
A sensor (I am not really sure where if it is best to call this sensor/camera) interface for interacting with any camera via OpenCV. It follows the same logic as the robot_interface. 

### Details
- The observation structure consists of an image array and the timestamp at which it was grabbed from the video stream.
- Can get the latest observation or observations at some other time indices, where these time indices can be queried too.
- Right now, the interface is only compatible with camera devices, in particular, those that can be interacted with using OpenCV.

## How I Tested
Via demo_camera.py by both visualizing the image stream and inspecting the pickled objects in which the images are stored.


include/robot_interfaces/opencv_driver.hpp
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @return Image frame consisting of an image matrix and the time at
+     * which it was grabbed.
+     */
+
+    OpenCVObservation grab_frame()
+    {
+        OpenCVObservation image_frame;
+        cv::Mat frame;
+        long double current_time = timer.get_current_time_sec();
```
luator: `get_current_time_sec` is a static method so you don't need an actual `Timer` object.  Instead you can call it like this: `real_time_tools::Timer::get_current_time_sec()`.

Since capturing the frame may take some time (e.g. because we need to wait for the camera to actually record the next frame), it would be better to get the timestamp afterwards, not before.

Further this function returns a `double`, not a `long double`.

include/robot_interfaces/opencv_driver.hpp
```diff
@@ -0,0 +1,73 @@
...
+        {
+#ifdef VERBOSE
+            std::cout << "Succeeded in accessing camera stream!" << std::endl;
+#endif
+            return 1;
+        }
+    }
+
+    /**
+     * @brief Grab frames along with their timestamps one by one.
```
luator: This method only grabs a single frame, so the description is a bit misleading.

include/robot_interfaces/opencv_driver.hpp
```diff
@@ -0,0 +1,73 @@
...
+class OpenCVDriver
+{
+public:
+    cv::VideoCapture video_capture;
+    real_time_tools::Timer timer;
+
+    OpenCVDriver()
+    {
+        cv::VideoCapture cap(0);
+        video_capture = cap;
```
luator: Better initialize `video_capture` directly with the desired argument:
```suggestion
    OpenCVDriver():
        video_capture(0)
    {
```

To be a bit more flexible, it might be nice to pass the argument to the constructor (can have `0` as default value). This would be relevant when multiple cameras are connected.

include/robot_interfaces/opencv_driver.hpp
```diff
@@ -0,0 +1,73 @@
...
+{
+/**
+ * @brief Driver for interacting with any camera using OpenCV.
+ */
+
+template <typename OpenCVObservation>
+class OpenCVDriver
+{
+public:
+    cv::VideoCapture video_capture;
```
luator: In C++ class variables need to be suffixed with `_`.
```suggestion
    cv::VideoCapture video_capture_;
```

include/robot_interfaces/opencv_driver.hpp
```diff
@@ -0,0 +1,73 @@
...
+        cv::VideoCapture cap(0);
+        video_capture = cap;
+    }
+
+    /**
+     * @brief Find out if the camera can be accessed and
+     * if the video capture has been started.
+     */
+
+    int is_grabbing_successful()
```
luator: Better return a bool value here, this would make it more obvious what values are to be expected.

include/robot_interfaces/pybind_camera.hpp
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+template <typename Types>
+void create_camera_bindings(pybind11::module& m)
+{
+    pybind11::class_<typename Types::Data, typename Types::DataPtr>(
+        m, "SensorData")
+        .def(pybind11::init<>());
+
+    pybind11::class_<typename Types::Backend>(m, "SensorBackend")
```
luator: To be a bit more compact and consistent with the robot types, remove the "Sensor" from the Python type.  It will be clear from the namespace what kind of backend it is (e.g. we have `robot_interfaces.finger.Backend` and `robot_interfaces.trifinger.Backend`).

include/robot_interfaces/pybind_camera.hpp
```diff
@@ -0,0 +1,69 @@
...
+        .def("grab_frame", &Types::CVDriver::grab_frame);
+
+    pybind11::class_<typename Types::Frontend>(m, "SensorFrontend")
+        .def(pybind11::init<typename Types::DataPtr>())
+        .def("get_latest_observation", &Types::Frontend::get_latest_observation)
+        .def("get_observation", &Types::Frontend::get_observation)
+        .def("get_timestamp_ms", &Types::Frontend::get_timestamp_ms)
+        .def("get_current_timeindex", &Types::Frontend::get_current_timeindex);
+
+    pybind11::class_<typename Types::OpenCVObservation>(m, "OpenCVObservation")
```
luator: Analog to what I wrote above, this should just be `Observation`.

include/robot_interfaces/sensor_backend.hpp
```diff
@@ -0,0 +1,92 @@
...
+namespace robot_interfaces
+{
+/**
+ * @brief Communication link between SensorData and SensorDriver.
+ *
+ * At each instant, it checks if the camera can be accessed and
+ * the video capture is running, if yes, then grabs an image and
+ * stores it along with the timestamp at which it was grabbed.
+ *
+ * @tparam OpenCVObservation
```
luator: The template argument should just be `Observation`.  `OpenCVObservation` would be a possible value to pass.

include/robot_interfaces/sensor_backend.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+    SensorBackend(
+        std::shared_ptr<OpenCVDriver<OpenCVObservation>> opencv_driver,
+        std::shared_ptr<SensorData<OpenCVObservation>> sensor_data)
+        : opencv_driver_(opencv_driver),
+          sensor_data_(sensor_data),
+          destructor_was_called_(false)
+    {
+        thread_ = std::make_shared<real_time_tools::RealTimeThread>();
+        thread_->create_realtime_thread(&SensorBackend::loop, this);
```
luator: I am not sure if we should use a real-time thread here.  Access to the camera will anyway not be real-time.

include/robot_interfaces/sensor_backend.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+    std::vector<real_time_tools::Timer> timers_;
+
+    std::shared_ptr<real_time_tools::RealTimeThread> thread_;
+
+    static void *loop(void *instance_pointer)
+    {
+        ((SensorBackend *)(instance_pointer))->loop();
+        return nullptr;
+    }
```
luator: Instead of defining this static function here, a lambda function can be used when creating the thread above.
See [here](https://github.com/open-dynamic-robot-initiative/blmc_robots/blob/bd9b010076286118a81648b9184d28ec5d9b192a/include/blmc_robots/n_joint_blmc_robot_driver.hpp#L642) for an example.

include/robot_interfaces/sensor_backend.hpp
```diff
@@ -0,0 +1,92 @@
...
+        thread_->join();
+    }
+
+private:
+    std::shared_ptr<OpenCVDriver<OpenCVObservation>> opencv_driver_;
+    std::shared_ptr<SensorData<OpenCVObservation>> sensor_data_;
+
+    bool destructor_was_called_;
+
+    std::vector<real_time_tools::Timer> timers_;
```
luator: unused?

include/robot_interfaces/sensor_backend.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * @brief Main loop.
+     */
+
+    void loop()
+    {
+        for (long int t = 0; !destructor_was_called_; t++)
+        {
+            OpenCVObservation camera_observation;
```
luator: Don't use "camera_" here.

include/robot_interfaces/sensor_backend.hpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    virtual ~SensorBackend()
+    {
+        destructor_was_called_ = true;
+        thread_->join();
+    }
+
+private:
+    std::shared_ptr<OpenCVDriver<OpenCVObservation>> opencv_driver_;
```
luator: This should be independent of the actual sensor, so don't use "opencv_driver_" here.  For this to work, we well need an abstract base class `SensorDriver` from which the actual driver classes inherit (analog to `RobotDriver`).

include/robot_interfaces/sensor_backend.hpp
```diff
@@ -0,0 +1,92 @@
...
+    void loop()
+    {
+        for (long int t = 0; !destructor_was_called_; t++)
+        {
+            OpenCVObservation camera_observation;
+
+            int flag = opencv_driver_->is_grabbing_successful();
+            if (flag == 1)
+            {
+                camera_observation = (opencv_driver_->grab_frame());
```
luator: Unnecessary parentheses.

include/robot_interfaces/sensor_backend.hpp
luator: Minor detail but it is sufficient to define this variable inside the `if` as it is only used there.

include/robot_interfaces/opencv_driver.hpp
```diff
@@ -0,0 +1,73 @@
+#pragma once
+
+#include <ctime>
+#include <iostream>
+
+#include <real_time_tools/process_manager.hpp>
+#include <real_time_tools/thread.hpp>
+#include <real_time_tools/threadsafe/threadsafe_timeseries.hpp>
```
luator: I think these includes are unused

include/robot_interfaces/sensor_data.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <iostream>
+#include <memory>
+#include <string>
+
+#include <time_series/time_series.hpp>
+
+namespace robot_interfaces
+{
+template <typename Type>
+using Timeseries = time_series::TimeSeries<Type>;
```
luator: Better move this inside the class to avoid collisions in the robot_interfaces namespace.

include/robot_interfaces/sensor_data.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @brief Contains all the data coming from the sensors.
+ *
+ * @tparam OpenCVObservation
+ */
+
+template <typename OpenCVObservation>
```
luator:
```suggestion
template <typename Observation>
```

include/robot_interfaces/sensor_data_types.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * @brief Collection of data types for the sensors used in the trifinger
+ * platform.
+ *
+ * Refer to the docstrings of individual data types for more details.
+ */
+
+struct SensorDataTypes
+{
+    typedef cv::Mat Image;
+    typedef long double TimeStamp;
```
luator: As mentioned above, better use only a normal `double`.

include/robot_interfaces/sensor_data_types.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace robot_interfaces
+{
+/**
+ * @brief Collection of data types for the sensors used in the trifinger
+ * platform.
+ *
+ * Refer to the docstrings of individual data types for more details.
+ */
+
+struct SensorDataTypes
```
luator: As this is specifying the types for the OpenCV camera, it should rather be called `OpenCVCameraTypes` or maybe better just `CameraTypes` as any other camera driver would probably also use cv::Mat to store the frame, no matter how it is grabbed.

luator: After thinking a bit more about it, I am actually unsure if we even need this struct at all. We could instead just define the Observation type and then specify the template argument everywhere when creating the frontend, backend, etc.

include/robot_interfaces/sensor_frontend.hpp
```diff
@@ -0,0 +1,58 @@
...
+namespace robot_interfaces
+{
+template <typename Type>
+using Timeseries = time_series::TimeSeries<Type>;
+typedef time_series::Index TimeIndex;
+
+/**
+ * @brief Communication link between SensorData and the user.
+ *
+ * Exposes the sensor data to the user via hardware-friendly methods.
```
luator: What do you mean by "hardware-friendly"?

include/robot_interfaces/sensor_frontend.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <time_series/time_series.hpp>
+
+#include <robot_interfaces/sensor_data.hpp>
+
+namespace robot_interfaces
+{
+template <typename Type>
+using Timeseries = time_series::TimeSeries<Type>;
+typedef time_series::Index TimeIndex;
```
luator: Like above, I think it is safer to define these inside the class.

include/robot_interfaces/sensor_frontend.hpp
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    OpenCVObservation get_observation(const TimeIndex &t)
+    {
+        return (*sensor_data_->observation)[t];
+    }
+
+    OpenCVObservation get_latest_observation()
+    {
+        return (*sensor_data_->observation)[get_current_timeindex()];
```
luator: TimeSeries has a method `newest_element` that you can use:
```suggestion
        return sensor_data_->observation->newest_element();
```

readme.md
```diff
@@ -18,4 +18,54 @@ Copyright(c) 2018-2019 Max Planck Gesellschaft
 
 BSD 3-Clause License
 
+## Steps to set up the pylon interface
```
luator: Pylon-related things should be removed from this PR.

readme.md
```diff
@@ -18,4 +18,54 @@ Copyright(c) 2018-2019 Max Planck Gesellschaft
...
+Just connect to a USB 3 port and install guvcview following the instructions [here](http://ubuntuhandbook.org/index.php/2018/10/guvcview-2-0-6-released-install-ubuntu/).
+
+### General productivity instructions
+
+Touch a CATKIN_IGNORE in the following packages-
+       * dbot
+       * dbot_ros
+       * fl
+       * dbot_ros_msgs
+       * blmc_ros_msgs
```
luator: Don't put this in the readme.

readme.md
```diff
@@ -18,4 +18,54 @@ Copyright(c) 2018-2019 Max Planck Gesellschaft
...
+
+### Steps to access pylon from inside the container
+
+```bash
+singularity shell -B /opt/pylon5 containers/blmc_ei.sif
+```
+
+### Steps to setup the webcam
+
+Just connect to a USB 3 port and install guvcview following the instructions [here](http://ubuntuhandbook.org/index.php/2018/10/guvcview-2-0-6-released-install-ubuntu/).
```
luator: Why would this be needed?

demos/demo_camera.py
```diff
@@ -0,0 +1,28 @@
...
+
+num_imgs = 100000000
+img_list = []
+
+for i in range(num_imgs):
+    OpenCVObservation = (camera_frontend.get_latest_observation())
+    img_list.append(OpenCVObservation)
+    print("At image", i)
+    window_name = 'image'
+    cv2.imshow(window_name, np.array(OpenCVObservation.image))
```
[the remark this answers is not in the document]
shrutij01: The blog post mentioned it's a conversion to np.ndarray but it doesn't seem like that to me since I needed to explicitly convert is to an np array. And it outputs the type as the name that's passes to pybind11::class_<cv::Mat>.

include/robot_interfaces/opencv_driver.hpp
shrutij01: Okay.

But the timestamp is a `long double`.  I had defined the time_stamp field of the observation to be a `long double Timstamp`. Should I change that?

include/robot_interfaces/sensor_frontend.hpp
shrutij01: Thanks

include/robot_interfaces/sensor_frontend.hpp
shrutij01: just to emphasize the point of the frontend (that is the methods the users can use won't break anything)

include/robot_interfaces/opencv_driver.hpp
shrutij01: Ok, understood. I had mistaken the type to be the same for the TimeStamp from `time_series` and from `real_time_tools`.

include/robot_interfaces/sensor_data_types.hpp
shrutij01: Yeah I had thought about this too, then just stuck to the same structure. Changed now, thanks.

include/robot_interfaces/sensor_frontend.hpp
luator: The frontend does not really do anything from preventing the user to break the hardware, it just ensures they do not mess with the data in the time series.  The safety checks are happening in the backend (but anyway not for the sensor, as we don't send commands to it).

include/robot_interfaces/opencv_driver.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <opencv2/opencv.hpp>
+
+namespace robot_interfaces
+{
+/**
+ * @brief Driver for interacting with any camera using OpenCV.
+ */
+
+template <typename OpenCVObservation>
+class OpenCVDriver
```
luator: I just noticed that the driver should actually not be templated as it needs that specific observation type.

demos/demo_camera.py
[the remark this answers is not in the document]
luator: At the bottom of the blog he actually also converts to `np.array` manually.  However, he adds a `copy=False` argument to it, to avoid copy of the data:
```
def get_image():
    im = myextension.get_image()
    return np.array(im, copy=False)
```
From what I understood `np.asarray(im)` should be the same as `np.array(im, copy=False)`, so I would suggest to use that in our code.

shrutij01:
@luator I think I made all the requested changes (except for the lamda function and non-realtime thread)

CMakeLists.txt
```diff
@@ -29,6 +29,17 @@ set(CATKIN_PKGS ${CATKIN_PKGS}
...
+find_package(Pylon REQUIRED)
+
+get_cmake_property(_variableNames VARIABLES)
+list (SORT _variableNames)
+foreach (_variableName ${_variableNames})
+    message(STATUS "${_variableName}=${${_variableName}}")
+endforeach()
+
+# include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindPylon.cmake")
+# message(FATAL_ERROR "XXXXXXXXXXXXX ${PYLON_INCLUDE_DIR}")
```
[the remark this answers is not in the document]
luator: I just noticed that the part for printing the variables is still there.

luator:
@shrutij01 There are still a few open issues, see above.
Apart from this, everything looks good.

CMakeLists.txt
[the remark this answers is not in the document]
shrutij01: Oh yes, I am sorry, I had fixed this but then was shifting this and FindPylon to another branch and made a mistake there. Fixed now. Sorry.

shrutij01:
@luator I fixed the unresolved threads now :)

PR #41, opened 2020-03-13
Accessing a single camera using the Pylon interface

shrutij01:
## Description

Adds a pylon driver to interact with a camera through the Basler pylon Camera Software Suite. The images are stored in a general CameraObservation structure which is common across cameras that require pylon and those which do not. Can switch between using/not using pylon by just specifying which driver to create (Pylon/OpenCV). 

## Note

1. Reducing the scope of pylon_image_ results in a segmentation fault. 
2. Pylon::PylonAutoInitTerm auto_init_term_; or else results in "Caught an unknown exception!"
3. Have kept the is_access_successful method only as a dummy since pylon throws exceptions anyway.

## How I Tested

Through the demo_camera.py by streaming the camera output.

- [ x] All new code is formatted according to our style guide (for C++ run clang-format, for Python, run flake8 and fix all warnings).
- [ x] All new functions/classes are documented and existing documentation is updated according to changes.
- [x ] No commented code from testing/debugging is kept (unless there is a good reason to keep it).


CMakeLists.txt
```diff
@@ -29,7 +29,8 @@ set(CATKIN_PKGS ${CATKIN_PKGS}
   time_series
 )
 find_package(catkin REQUIRED COMPONENTS ${CATKIN_PKGS})
-find_package(OpenCV 3 REQUIRED)
+find_package(OpenCV 3)
+find_package(Pylon REQUIRED)
```
luator: Pylon should be an optional dependency, otherwise everyone needs to install it.

If it is not found, the variable `${PYLON_FOUND}` will not be set (or set to false).  Add the following lines to be able to access this variable in the C++ code:

    if ${PYLON_FOUND}
        add_definitions(-DPYLON_FOUND)
    endif()

Then wrap all usages of Pylon in C++ with

    #ifdef PYLON_FOUND
       ...
    #endif

I think here this would only be the include and the python-binding of the pylon driver in `py_camera_types.cpp`

cmake/FindPylon.cmake
```diff
@@ -0,0 +1,159 @@
+#############################################################################
```
luator: I'll try to figure out if adding this file affects the license of our other files.

luator: I just found [this one](https://github.com/basler/pylon-ros-camera/blob/master/pylon_camera/cmake/FindPylon.cmake), can you give it a try?

It is part of the ROS package provided by Basler, so I would assume it should work.  The package is licensed with BSD, so we can use this without headache.

In case you use this file, please add a comment header with a link to the original file and simply put the complete content of [this LICENSE file](https://github.com/basler/pylon-ros-camera/blob/master/pylon_camera/LICENSE) there as well.

shrutij01: This cmake file's not working. I will look into the variable names, but ideally this should just work, right?

shrutij01: Ok I changed all "Pylon"s to "PYLON"s and it works now

CMakeLists.txt
shrutij01: Fixed this, thanks

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -30,15 +30,28 @@ class PylonDriver : public SensorDriver<CameraObservation>
 public:
     Pylon::PylonAutoInitTerm auto_init_term_;
     Pylon::CInstantCamera camera_;
+    Pylon::CTlFactory& tl_factory_;
+    Pylon::DeviceInfoList_t device_list_;
+    Pylon::DeviceInfoList_t::const_iterator device_iterator_;
```
luator: It seems that these variables are only used in the constructor? If so, please move them there.

cmake/FindPylon.cmake
luator: I think it would be better to do it the other way round, i.e. keep the FindPylon.cmake as it is and change to `Pylon_` in our CMakeLists.txt.  This way it would be more compliant with CMake conventions.

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -31,27 +31,66 @@ class PylonDriver : public SensorDriver<CameraObservation>
...
+
+            for(device_iterator = device_list.begin(); device_iterator != device_list.end(); ++device_iterator)
+            {
+                std::string device_user_id_found(device_iterator->GetUserDefinedName());
+                if((0 == device_user_id_to_open.compare(device_user_id_found)) ||
+                (device_user_id_to_open.length() < device_user_id_found.length() &&
+                (0 == device_user_id_found.compare(device_user_id_found.length() -
+                                                    device_user_id_to_open.length(),
+                                                    device_user_id_to_open.length(),
+                                                    device_user_id_to_open) )
```
luator: I know this comes from the ROS node, I think a simple `device_user_id_found == device_user_id_to_open` would be sufficient, though.
This code seems to match also if the given id is shorter but matches the beginning of the found one. I don't think we need/want this feature.  So if a simple `==` works, I would just use that.

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -31,27 +31,66 @@ class PylonDriver : public SensorDriver<CameraObservation>
...
+
+            if (found_desired_device)
+            {
+                camera_.Attach(tl_factory_.CreateDevice(*device_iterator));
+            }
+            else
+            {
+                device_iterator = device_list.begin();
+                camera_.Attach(tl_factory_.CreateDevice(*device_iterator));
+                std::cout << "Desired device not found. Creating a camera object with the first device id in device list." << std::endl;
```
luator: I would rather throw an exception if the desired camera is not found.
However, I think it would be nice to add a case like this (i.e. using the first camera) in case the user sets an empty string as id.

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -88,11 +127,16 @@ class PylonDriver : public SensorDriver<CameraObservation>
                                             CV_8UC3,
                                             (uint8_t*)pylon_image_.GetBuffer());
             }
+            else
+            {
+                throw "Failed to access images from the camera.";
```
luator: Better throw an exception than just a string
```suggestion
                throw std::runtime_error("Failed to access images from the camera.");
```

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -31,27 +31,66 @@ class PylonDriver : public SensorDriver<CameraObservation>
     Pylon::PylonAutoInitTerm auto_init_term_;
     Pylon::CInstantCamera camera_;
     Pylon::CTlFactory& tl_factory_;
-    Pylon::DeviceInfoList_t device_list_;
-    Pylon::DeviceInfoList_t::const_iterator device_iterator_;
     Pylon::CImageFormatConverter format_converter_;
     Pylon::CPylonImage pylon_image_;
-    int num_devices_;
+    const std::string& device_user_id_to_open = "cam_1";
```
luator: Better pass this as argument to the constructor.

cmake/FindPylon.cmake
shrutij01: I changed it PYLON in the find_package then, that's fine right?

shrutij01: Do we need to follow the CMake convention? I was just thinking it's cleaner now since it's in caps (as PYLON) everywhere except in find_package() where it's "Pylon". Otherwise it's a mix of both everywhere.

shrutij01:
@luator I think I have fixed almost all issues except for,
- PixelType as just storing as a BayerBG8 results in an unknown exception when converting to a cv::Mat with type CV_8UC3, will check for the right non-3 channel type).
- The naming convention in CMake. I let Pylon remain thus in find_package() but changed to PYLON in the rest of the places as before it was just a mix of both everywhere, which looked much less cleaner.

demos/demo_camera.py
```diff
@@ -17,15 +17,15 @@ def main():
...
     arg_action_group.add_argument("--pylon",
-                                  action="store_false",
+                                  action="store_true",
                                   help="""Pass --pylon if you need to access
                                   the camera via pylon.
                                   """)
     args = argparser.parse_args()
 
     camera_data = robot_interfaces.camera.Data()
     if args.pylon == 1:
```
luator: With the change above, `args.pylon` is a bool, so `if args.pylon:` is enough.

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -129,15 +133,13 @@ class PylonDriver : public SensorDriver<CameraObservation>
...
             {
-                throw "Failed to access images from the camera.";
+                throw std::runtime_error(
+                    "Failed to access images from the camera.");
             }
-            
-            
         }
-        catch (const char* error_msg)
+        catch (std::runtime_error& e)
```
luator: exceptions shouldn't be caught inside this function but in the backend loop

cmake/FindPylon.cmake
luator: Yes, we should follow these conventions. It makes the code much easier to use because you don't need to dig and figure out how the variables are named.

What I meant is to change it to `Pylon_` everywhere, so there would be no mix.

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -33,58 +33,71 @@ class PylonDriver : public SensorDriver<CameraObservation>
     Pylon::CTlFactory& tl_factory_;
     Pylon::CImageFormatConverter format_converter_;
     Pylon::CPylonImage pylon_image_;
-    const std::string& device_user_id_to_open = "cam_1";
+    const std::string& device_user_id_to_open_;
```
luator: These variables are all only used inside single functions, so I think they could be defined locally.

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -33,58 +33,71 @@ class PylonDriver : public SensorDriver<CameraObservation>
...
+                if (found_desired_device)
+                {
+                    camera_.Attach(tl_factory_.CreateDevice(*device_iterator));
+                }
+                else
+                {
+                    throw std::runtime_error(
+                        "Device id specified doesn't correspond to any "
+                        "connected devices, please retry with a valid id.");
+                    Pylon::PylonTerminate();
```
luator: This should be moved before the `throw`, otherwise it will never be executed.

cmake/FindPylon.cmake
```diff
@@ -0,0 +1,60 @@
...
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+
```
luator: Please add a link to the original file

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -0,0 +1,148 @@
...
+namespace robot_interfaces
+{
+/**
+ * @brief Driver for interacting with a camera via Pylon and store
+ * images using OpenCV.
+ */
+class PylonDriver : public SensorDriver<CameraObservation>
+{
+public:
+    Pylon::PylonAutoInitTerm auto_init_term_;
```
luator: I was curious about what this thing does.  From what I understood, it takes care of initializing and terminating, so it should not be necessary to explicitly call `PylonInitialize` and `PylonTerminate` below.

luator:
Regarding pixel type: The proper type would probably be `CV_8UC1`, but I would suggest to keep BGR for now an take care of this in a later PR.

include/robot_interfaces/sensors/pylon_driver.hpp
shrutij01: but they are being called by the constructor, so shouldn't they be defined as class variables?

shrutij01: Not sure if I am missing something, but const and reference members cannot be initialized in the constructor, right?

include/robot_interfaces/sensors/pylon_driver.hpp
shrutij01: Yeah, it takes care of initializing and terminating. I added the PylonTerminate explicitly to terminate in case of runtime_error exceptions like when there is no device connected, or when the grab cannot succeed. So, added PylonInitialize to go with it. Should I remove them then?

shrutij01:
@luator I have implemented the recommended fixes, but am just not sure about the PylonInitialize and PylonTerminate. Let me know if I should remove the explicit PylonTerminate from the checks, and hence the PylonInitialize too.

include/robot_interfaces/sensors/pylon_driver.hpp
luator: If the explicit terminate before throwing is needed, then go for that method.  I think in this case you can simply remove the "Auto" thing, because from what I saw you already call Initialize/Terminate everywhere properly.

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -116,30 +124,23 @@ class PylonDriver : public SensorDriver<CameraObservation>
...
+            image_frame.image = cv::Mat(ptr_grab_result->GetHeight(),
+                                        ptr_grab_result->GetWidth(),
+                                        CV_8UC3,
+                                        (uint8_t*)pylon_image_.GetBuffer());
         }
-        catch (std::runtime_error& e)
+        else
         {
-            std::cerr << e.what() << std::endl;
+            Pylon::PylonTerminate();
```
luator: I am not sure if terminating here (inside `get_observation`) is good.  It might be that there was just an issue with this frame, so the backend could just skip it an continue trying to get observations.  However, when Pylon is terminated here, the camera will probably stop working completely.  So I think it is better to rely on the backend to catch this exception and not terminate here.

shrutij01: Ah yes, you're right. I had not thought about that. Will fix this, thank you :)

include/robot_interfaces/sensors/pylon_driver.hpp
shrutij01: Hmm yeah, I tried doing this now, but removing the auto init results in a runtime error of "Caught an unknown exception!". I am not sure what this auto is really doing

luator: Okay, then just leave it there.  I guess as long as everything works, that's good enough for us :)

shrutij01: ok :)

shrutij01:
I just added a line to the readme so that someone does not accidentally use a different camera (in case they're not paying attention to the view (which I have been guilty of myself :3)).

include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -36,17 +36,14 @@ class PylonDriver : public SensorDriver<CameraObservation>
...
-     * @param device_user_id The id of the camera device to open and
+     * @param device_user_id_to_open The id of the camera device to open and
      * grab images from
      */
-    PylonDriver(const std::string& device_user_id)
-        : device_user_id_to_open_(device_user_id),
-          tl_factory_(Pylon::CTlFactory::GetInstance())
+    PylonDriver(const std::string& device_user_id_to_open)
     {
+        Pylon::CTlFactory& tl_factory_ = Pylon::CTlFactory::GetInstance();
```
luator: Minor thing, but please remove the trailing `_` to avoid confusion

shrutij01: Yep sorry had missed that, fixed now.

shrutij01:
@luator thank you so much! :D

PR #42, opened 2020-03-20
Release GIL when calling C++ methods, terminate backend cleanly & add real-time mode

luator:
## Description

By releasing the GIL while calling the (C++) frontend methods, other
threads can acquire it if needed.

This is, for example, needed for the pyBullet Driver where the backend
loop needs to call Python functions and needs the GIL for that.

## How I Tested
By running the pipeline with the pyBullet driver (to be provided in another PR).
Not tested with real robot due to current lack of access.

### DIFF
--- a/include/robot_interfaces/robot_backend.hpp
+++ b/include/robot_interfaces/robot_backend.hpp
@@ -222,7 +222,6 @@ private:
             {
                 std::cerr << "Error: " << status.error_message
                           << "\nRobot is shut down." << std::endl;
-                robot_driver_.shutdown();
                 break;
             }
             timer_.checkpoint("status");
@@ -234,7 +233,6 @@ private:
             }
             if (has_shutdown_request())
             {
-                // TODO should shut down robot?
                 break;
             }
 
@@ -253,6 +251,7 @@ private:
             }
         }
 
+        robot_driver_.shutdown();
         loop_is_running_ = false;
     }
 };
--- a/include/robot_interfaces/robot_backend.hpp
+++ b/include/robot_interfaces/robot_backend.hpp
@@ -47,14 +47,21 @@ public:
      * @param robot_data  Data is send to/retrieved from here.
      * @param max_action_duration_s  See MonitoredRobotDriver.
      * @param max_inter_action_duration_s  See MonitoredRobotDriver.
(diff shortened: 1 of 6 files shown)


luator:
I added several commits related to also releasing the GIL when running a backend-only script and to terminate the backend cleanly (i.e. react to Ctrl+C without haning anywhere).

luator:
I also added the `real_time_mode` option to the backend which is needed for the simulation.  It defaults to `true`, so it should not break existing code.

luator:
rebased to resolve merge conflict

include/robot_interfaces/robot_backend.hpp
```diff
@@ -47,18 +47,27 @@ class RobotBackend
...
-    RobotBackend(
-        std::shared_ptr<RobotDriver<Action, Observation>> robot_driver,
-        std::shared_ptr<RobotData<Action, Observation>> robot_data,
-        const double max_action_duration_s,
-        const double max_inter_action_duration_s)
+    RobotBackend(std::shared_ptr<RobotDriver<Action, Observation>> robot_driver,
+                 std::shared_ptr<RobotData<Action, Observation>> robot_data,
+                 const double max_action_duration_s,
+                 const double max_inter_action_duration_s,
+                 const bool real_time_mode = true)
```
mwuethri: i am worried that this is becoming slightly obscure, i.e. it is not obvious at first sight how these parameters interact. 

would it make sense to e.g. pass it directly a monitored driver, such that it does not need to construct it internally, and hence it would not need the arguments max_action_duration_s, max_inter_action_duration_s? this would encapsulate parameters more. also, i think it would be good to have max_action_repetitions as an argument here as well, then everything is together.

i will approve anyway, but maybe this would be nice for the future, because for somebody who is not used to it, i think this is a bit confusing.

mwuethri: in fact, if we pass it directly a monitored/unmonitored driver, and it does not need to do construction internally, then this safety check is abstracted away. for a simulator for instance, you don't need to monitor the driver, you can directly pass it the unmonitored driver. so i think this would make things much easier to understand, and it makes sense particularly also for the simulation driver.

luator: Sounds good, I will do this in a separate PR.

Issue #40, opened 2020-03-13
SensorBackend should use non-real-time thread


luator:
To my understanding, the cameras will anyway not work in real time, so there is probably no use of running the `SensorBackend` in a real-time thread. On the other hand, it is probably beneficial to keep the number of real-time threads low, so it would be better to use a standard non-real-time thread here.

https://github.com/open-dynamic-robot-initiative/robot_interfaces/blob/1fa5daa15c99d2a47e7d9f106e33bd8011d54127/include/robot_interfaces/sensors/sensor_backend.hpp#L49

shrutij01:
Yes, thanks, am looking into this.

PR #43, opened 2020-03-20
Changing to a non-realtime thread in the sensor backend

shrutij01:
## Description

Earlier, the sensor backend was creating a realtime thread for the loop getting observations through the sensor driver and appending these to the sensor data. But as this anyway does not run in realtime, a non-realtime thread is used now. 

This fixes issue #40.

## How I Tested

Through the demo_camera


shrutij01:
sorry i did not create this from master.

PR #44, opened 2020-03-20
Changing to a non-realtime thread in the sensor backend

### LINKED ISSUE #40: SensorBackend should use non-real-time thread

shrutij01:
## Description

Earlier, the sensor backend was creating a realtime thread for the loop getting observations through the sensor driver and appending these to the sensor data. But as this anyway does not run in realtime, a non-realtime thread is used now. 

This fixes #40.

## How I Tested

Through the demo_camera

### DIFF
--- a/include/robot_interfaces/sensors/sensor_backend.hpp
+++ b/include/robot_interfaces/sensors/sensor_backend.hpp
@@ -11,6 +11,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
+#include <thread>
 
 #include <real_time_tools/process_manager.hpp>
 #include <real_time_tools/thread.hpp>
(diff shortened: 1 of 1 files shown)


include/robot_interfaces/sensors/sensor_backend.hpp
```diff
@@ -11,6 +11,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
+#include <thread>
 
 #include <real_time_tools/process_manager.hpp>
 #include <real_time_tools/thread.hpp>
```
luator: these includes can now be removed.

PR #47, opened 2020-03-31
[WIP: Basic] Syncing Three pylon Dependent Cameras

shrutij01:
## Description

A very basic way of creating three instances of the PylonDriver and then testing how fast this can work. Adds the SyncDriver for this, which relies on the safety checks being performed in the PylonDriver itself. It derives from the SensorDriver with TricameraObservation as the ObservationType. This is mostly for structural sanity at this stage.

## How I Tested

Not tested yet. But basic check through the demo_tricamera.py


demos/demo_tricamera.py
```diff
@@ -0,0 +1,32 @@
...
+import cv2
+import pickle
+
+import robot_interfaces
+
+
+def main():
+
+    camera_data = robot_interfaces.tricamera.Data()
+    camera_driver = robot_interfaces.tricamera.SyncDriver("cam_1", "cam_2", "cam_3")
```
luator, line 16: For testing on the TriFinger platform: the cameras there are called "camera60", "camera180" and "camera300" (rougly refering to their angle like in the naming of the fingers)

include/robot_interfaces/sensors/sync_driver.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <robot_interfaces/sensors/sensor_driver.hpp>
+#include <robot_interfaces/sensors/tricamera_observation.hpp>
+
+namespace robot_interfaces
+{
+/**
+ * @brief Driver to create three instances of the PylonDriver
+ * and get observations from them.
+ */
+class SyncDriver : public SensorDriver<TricameraObservation>
```
luator, line 28: I think "TricameraDriver" would be more clear.

luator: Or better "Tri**C**ameraDriver, then it is consistent with the TriFinger

include/robot_interfaces/sensors/tricamera_observation.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+/**
+ * @brief Observation structure to store cv::Mat images with corresponding
+ * timestamps.
+ */
+struct TricameraObservation
+{
+    typedef std::array<CameraObservation, 3> CameraArray;
+    CameraArray cam_array;
+};
```
luator, line 30: At least as long as there is only this one element, we don't actually need a struct.  I think directly
```
typedef std::array<CameraObservation, 3> TricameraObservation;
```
should be enough.

luator:
The general structure looks good to me.  Let me know when you tested it successfully, then I will go through it in more detail.

shrutij01:
@luator, I have tested the driver. Attaching the distance between the time_stamps (I stored them as pickle, and then plotted to avoid additional lags). 
first_100
last_99

luator:
@shrutij01 What exactly are the plots showing?  Specifically, what are the different colours?

demos/demo_tricamera.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
```
luator, line 1: The file is not PEP8-compliant.  If you want to give `black` a try, limit the line length to 79, i.e.

    black --line-length 79 filename.py

demos/demo_tricamera.py
```diff
@@ -0,0 +1,42 @@
...
+        window_180 = "Image Stream camera180"
+        window_300 = "Image Stream camera300"
+        cv2.imshow(window_180, np.array(observation.cam_array[0].image, copy=False))
+        cv2.imshow(window_300, np.array(observation.cam_array[1].image, copy=False))
+        cv2.imshow(window_60, np.array(observation.cam_array[2].image, copy=False))
+        cv2.waitKey(3)
+
+        observations_timestamps_list.append([observation.cam_array[0].time_stamp, observation.cam_array[1].time_stamp, observation.cam_array[2].time_stamp])
+
+    pickle.dump(observations_timestamps_list, open("time_stamps.p", "wb"))
```
luator, line 36: I would suggest to add an argument to the script (using argparse) to specify the filename.  And don't save the file at all if no argument is given.

include/robot_interfaces/sensors/tricamera_driver.hpp
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+#pragma once
+
+#include <ctime>
+#include <iostream>
+
+#include <real_time_tools/timer.hpp>
+
+#include <pylon/PylonIncludes.h>
```
luator, line 17: These includes seem to be unused (from ctime to pylon)

include/robot_interfaces/sensors/tricamera_observation.hpp
luator: If keeping the struct, please give the `cam_array` a more meaningful name.  Not sure what's the best, maybe just `cameras`.

include/robot_interfaces/sensors/tricamera_observation.hpp
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+
+#include <cmath>
+#include <ctime>
+#include <real_time_tools/threadsafe/threadsafe_timeseries.hpp>
+#include <time_series/time_series.hpp>
+
+#include <robot_interfaces/sensors/camera_observation.hpp>
+
+#include <opencv2/opencv.hpp>
```
luator, line 18: Most of these includes are unused.

include/robot_interfaces/sensors/tricamera_observation.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace robot_interfaces
+{
+/**
+ * @brief Observation structure to store cv::Mat images with corresponding
+ * timestamps.
+ */
+struct TricameraObservation
```
luator, line 26: For consistency, please rename this to "Tri**C**ameraObservation".

shrutij01:
> @shrutij01 What exactly are the plots showing? Specifically, what are the different colours?

@luator I plotted the abs difference between the timestamps from the observations of a pair of cameras (so r, g, b correspond to the three pairs). This is for 100 observations. Attached both the plots since the maximum difference occurs for the first observation, and is much larger than the average difference from the second observation onwards.

demos/demo_tricamera.py
shrutij01: Thank you :) ! Sorry hadn't done this earlier.

demos/demo_tricamera.py
```diff
@@ -0,0 +1,68 @@
...
+import numpy as np
+import cv2
+import pickle
+
+import robot_interfaces
+
+
+def main():
+    argparser = argparse.ArgumentParser(description=__doc__)
+    arg_action_group = argparser.add_mutually_exclusive_group(required=False)
```
luator, line 17: There is no need for a mutually exclusive group if there is only one argument anyway.  You can use `argparser.add_argument`

demos/demo_tricamera.py
```diff
@@ -0,0 +1,68 @@
...
+        observations_timestamps_list.append(
+            [
+                observation.cameras[0].time_stamp,
+                observation.cameras[1].time_stamp,
+                observation.cameras[2].time_stamp,
+            ]
+        )
+
+    if args.store_timestamps:
+        pickle.dump(observations_timestamps_list, open("time_stamps.p", "wb"))
```
luator, line 64: You should use the value from the argument for the file name ;)

shrutij01:
@luator I think I have addressed all open issues now :)

srcpy/py_tricamera_types.cpp
```diff
@@ -0,0 +1,36 @@
...
+#endif
+
+#include <robot_interfaces/sensors/pybind_sensors.hpp>
+#include <robot_interfaces/sensors/sensor_driver.hpp>
+
+using namespace robot_interfaces;
+
+PYBIND11_MODULE(py_tricamera_types, m)
+{
+    create_sensor_bindings<TricameraObservation>(m);
```
luator, line 21: The struct is now defined as `TriCameraObservation` but here it is still with lower case "c"

srcpy/py_tricamera_types.cpp
```diff
@@ -0,0 +1,36 @@
...
+using namespace robot_interfaces;
+
+PYBIND11_MODULE(py_tricamera_types, m)
+{
+    create_sensor_bindings<TriCameraObservation>(m);
+
+#ifdef Pylon_FOUND
+    pybind11::class_<TriCameraDriver,
+                     std::shared_ptr<TriCameraDriver>,
+                     SensorDriver<TricameraObservation>>(m, "TriCameraDriver")
```
luator, line 26: It is still with lower-case "c" here.  Did you test compiling on a machine with Pylon installed?

shrutij01: Sorry about that, I had made a mistake in testing. Fixed now. Thanks for pointing this out.

shrutij01:
Thank you :D

PR #48, opened 2020-04-02
Fwidmaier/backend driver

luator:

## Description

Instead of always wrapping the robot driver in a monitored driver, just use the driver as it is given.  The user is now responsible to pass a monitored driver in case the monitoring is desired.  This has the advantage that for cases where no monitoring is desired (e.g. simulation), no wrapper with infinite timeouts needs to be created, but the actual driver can just be used directly.

With this change, the actual driver can just be used without the monitoring wrapper in case no monitoring is desired. So there should be no use-case for infinite timeouts anymore.

## How I Tested
By running a demo with the simulation backend.

## Do not merge before

Related PRs on other repos are ready.

- [ ] All new code is formatted according to our style guide (for C++ run clang-format, for Python, run flake8 and fix all warnings).
- [ ] All new functions/classes are documented and existing documentation is updated according to changes.
- [ ] No commented code from testing/debugging is kept (unless there is a good reason to keep it).


demos/demo.cpp
```diff
@@ -147,16 +147,7 @@ int main()
   std::shared_ptr<Driver> driver_ptr = std::make_shared<Driver>();
   std::shared_ptr<BaseData> data_ptr = std::make_shared<SingleProcessData>();
 
-  // max time allowed for the robot to apply an action.
```
luator, line 150: @mwuethri I am actually a bit unsure now.  In this demo no real robot is involved, so technically no monitoring is needed.  However, maybe it makes sense to keep it here for demonstration purposes?

luator:
@mwuethri I did some modifications:
- Fixed a merge conflict with master
- Simplified template arguments of the MonitoredRobotDriver.  In fact, with C++17, it should be possible to automatically deduct it now (unfortunately we cannot use this on Ubuntu 16.04 yet...).
- Do not remove the "is finite" check in the monitored driver, instead keep the current behaviour and print a warning if the loop is not started.

PR #51, opened 2020-04-08
RobotBackend: Add "First Action"-Timeout

luator:

## Description
Add a timeout to `RobotBackend` to shut down when too much time passes between creation of the backend and reception of the first action.
By default it is set to infinity, which disables it.

Also add a shutdown method to the backend which makes handling of the shutdown a bit easier.

## How I Tested

By modifying a demo and adding sleeps.

### DIFF
--- a/include/robot_interfaces/robot_backend.hpp
+++ b/include/robot_interfaces/robot_backend.hpp
@@ -58,7 +58,7 @@ public:
           robot_data_(robot_data),
           real_time_mode_(real_time_mode),
           first_action_timeout_(first_action_timeout),
-          shutdown_was_called_(false),
+          is_shutdown_requested_(false),
           max_action_repetitions_(0)
     {
         GlobalSignalHandler::initialize();
@@ -70,19 +70,8 @@ public:
 
     virtual ~RobotBackend()
     {
-        shutdown();
-    }
-
-    /**
-     * @brief Shut down the backend loop and wait until it terminated.
-     */
-    void shutdown()
-    {
-        if (!shutdown_was_called_)
-        {
-            shutdown_was_called_ = true;
-            thread_->join();
-        }
+        request_shutdown();
+        thread_->join();
     }
 
     uint32_t get_max_action_repetitions()
@@ -114,6 +103,17 @@ public:
     void initialize()
     {
         robot_driver_->initialize();
+    }
+
+    /**
+     * @brief Request shutdown of the backend loop.
+     *
+     * The loop may take some time to actually terminate after calling this
+     * function. Use wait_until_terminated() to ensure it has really terminated.
+     */
+    void request_shutdown()
+    {
+        is_shutdown_requested_ = true;
     }
 
     /**
@@ -156,12 +156,12 @@ private:
     const double first_action_timeout_;
 
     /**
-     * @brief Set to true when the destructor is called
+     * @brief Set to true when shutdown is requested.
      *
      * This is used to notify the background loop about requested shutdown, so
      * it terminates itself.
      */
-    std::atomic<bool> shutdown_was_called_;
+    std::atomic<bool> is_shutdown_requested_;
 
     //! @brief Indicates if the background loop is still running.
     std::atomic<bool> loop_is_running_;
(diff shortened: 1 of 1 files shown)


include/robot_interfaces/robot_backend.hpp
```diff
@@ -177,11 +201,29 @@ class RobotBackend
     {
         real_time_tools::set_cpu_dma_latency(0);
```
mwuethri, line 202: side note: i think we could get rid of this, as by default this is done when starting a realtime thread, right?

include/robot_interfaces/robot_backend.hpp
```diff
@@ -177,11 +201,29 @@ class RobotBackend
...
+                Status status;
+                status.error_status = Status::ErrorStatus::BACKEND_ERROR;
+                status.error_message = "First action was not provided in time";
+
+                robot_data_->status->append(status);
+
+                std::cerr << "Error: " << status.error_message
+                          << "\nRobot is shut down." << std::endl;
+
+                this->shutdown();
```
mwuethri, line 224: doesn't this wait forever? inside shutdown you join the thread, but here you are inside of that thread.. am i missing something?

luator: Ohh.. you are right...  I wonder why this was not happening when I tested.  Will change it in any case.

include/robot_interfaces/robot_backend.hpp
luator: I just checked, yes this is already done in `create_realtime_thread`.  I'll remove it everywhere but will do so in a separate PR.

include/robot_interfaces/robot_backend.hpp
luator: So it does not block but throws an `std::system_error("Resource deadlock avoided")`.  I did not check carefully enough earlier so I did not notice it was not the expected exception.  I should probably add a unit test for this.

luator:
- Regarding `loop_is_running_`: When adding this, I originally had a non-blocking `is_running()` method in mind.  Given that the only use is now in `wait_until_terminated()`, I agree I could just use `join()` there.  This requires some additional testing, though, so I'll do that in a separate PR.
- Regarding initialization: Right now, there is a check in the driver of the real robot, which raises an error when an action is applied before initialization.  It might make sense to move this check to the backend.  Or maybe replace `initizalize` with a `start` method and only start the loop there?  I think it is important to keep it as an external function call because it will cause the robot to move (due to homing), which the user should have control over.

include/robot_interfaces/robot_backend.hpp
luator: I replaced it with a non-blocking `request_shutdown` now.

mwuethri:
> * Regarding `loop_is_running_`: When adding this, I originally had a non-blocking `is_running()` method in mind.  Given that the only use is now in `wait_until_terminated()`, I agree I could just use `join()` there.  This requires some additional testing, though, so I'll do that in a separate PR.
> * Regarding initialization: Right now, there is a check in the driver of the real robot, which raises an error when an action is applied before initialization.  It might make sense to move this check to the backend.  Or maybe replace `initizalize` with a `start` method and only start the loop there?  I think it is important to keep it as an external function call because it will cause the robot to move (due to homing), which the user should have control over.

oh right, ok that makes sense then, i don't think we need to handle this in the backend then

include/robot_interfaces/robot_backend.hpp
```diff
@@ -177,11 +201,29 @@ class RobotBackend
...
+                Status status;
+                status.error_status = Status::ErrorStatus::BACKEND_ERROR;
+                status.error_message = "First action was not provided in time";
+
+                robot_data_->status->append(status);
+
+                std::cerr << "Error: " << status.error_message
+                          << "\nRobot is shut down." << std::endl;
+
+                this->request_shutdown();
```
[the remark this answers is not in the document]
luator: Previously, when the method was called `shutdown`, I thought it's good to be very explicit to avoid confusion with `robot_driver_->shutdown()`.  Now that it is renamed, it does not really make sense anymore, I'm going to remove it.

PR #57, opened 2020-04-27
Explicitly set some camera settings

luator:

## Description

This is an intermediate step of setting the camera configuration from our software.
This only sets a few things for now (exposure time, frame rate and white balance).  The next step is to have a proper configuration file with all settings but this will be done in a later PR.

## How I Tested
By running it on the platform.

- [ ] All new code is formatted according to our style guide (for C++ run clang-format, for Python, run flake8 and fix all warnings).
- [ ] All new functions/classes are documented and existing documentation is updated according to changes.
- [ ] No commented code from testing/debugging is kept (unless there is a good reason to keep it).

### DIFF
--- a/include/robot_interfaces/sensors/pylon_driver.hpp
+++ b/include/robot_interfaces/sensors/pylon_driver.hpp
@@ -31,6 +31,7 @@ namespace robot_interfaces
  */
 class PylonDriver : public SensorDriver<CameraObservation>
 {
+// TODO implementations should be moved to cpp file
 public:
     Pylon::PylonAutoInitTerm auto_init_term_;
     Pylon::CInstantCamera camera_;
@@ -99,6 +100,8 @@ public:
                 format_converter_.OutputPixelFormat =
                     Pylon::PixelType_BGR8packed;
 
+                set_camera_configuration(camera_.GetNodeMap());
+
                 camera_.StartGrabbing();
             }
         }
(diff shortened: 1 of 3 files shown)


include/robot_interfaces/sensors/pylon_driver.hpp
```diff
@@ -139,6 +142,25 @@ class PylonDriver : public SensorDriver<CameraObservation>
...
+        Pylon::CFloatParameter exposure_time(nodemap, "ExposureTime");
+        exposure_time.SetValue(1500);
+
+        //Pylon::CBooleanParameter enable_frame_rate(nodemap, "EnableAcquisitionFrameRate");
+        //enable_frame_rate.SetValue(true);
+
+        Pylon::CFloatParameter frame_rate(nodemap, "AcquisitionFrameRate");
+        frame_rate.SetValue(100);
+
+        Pylon::CEnumParameter(nodemap, "BalanceWhiteAuto").SetValue("Once");
```
shrutij01, line 158: Do we want to set the white balance only once? I just remembered we were setting it adaptively when we were viewing the stream from the PylonViewerApp.

luator: As the light conditions in the box should be constant, I think there is no need for permanent adjustment.
Originally I wanted to disable auto-white balance completely but unfortunately the same value leads to slightly different results for the different cameras.

luator:
Yes, the comments are for future use.  It's a bit quick and dirty but I plan to directly continue working on it.

PR #60, opened 2020-05-14
Add basic sensor logger

luator:

## Description
Add a simple `SensorLogger` that buffers sensor observations in memory and provides a method to write them to a file.

For writing to file, serialization with cereal is used, which makes this very easy (serialization functions for the observation types need to be provided anyway for the shared memory).

The data is currently buffered/saved without any compression, so the required memory can be very high.  If this turns out to be a problem, we need to think about some way of compressing the data.

As the resulting file format (binary serialization output of cereal) is nothing that can be opened with standard tools, there is also a simple `SensorLogReader` class through which users can easily access the data.

## How I Tested

Using the unit test added in this PR.


include/robot_interfaces/sensors/sensor_logger.hpp
```diff
@@ -0,0 +1,162 @@
...
+    DataPtr sensor_data_;
+    std::vector<Observation> buffer_;
+    size_t buffer_limit_;
+    std::thread buffer_thread_;
+    bool enabled_;
+
+    //! Get observations from sensor_data_ and add them to the buffer.
+    void loop()
+    {
+        auto t = sensor_data_->observation->newest_timeindex();
```
mwuethri, line 135: this could possibly get stuck here if no observation is ever added.

luator: Oh, I missed that.  I'll add a wait-loop in the beginning like in the other modules.
Or maybe it would be better to finally take the time to make the time series detect when the application is shut down and stop waiting.

luator: I'll anyway merge this now, to not have this PR pending for too long.

PR #68, opened 2020-06-23
Make methods of SensorFrontend const

luator:
## Description

- Make methods of SensorFrontend const (needed so they can be called in other const functions).
- Don't use references for time index.  It is just an integer, so there is no benefit from using a const reference.

## How I Tested

Compiled.


shrutij01:
lgtm. sorry, had forgotten to check earlier.

PR #72, opened 2020-07-17
Refactor Logger and add method to log without background thread

luator:
## Description

- Some refactoring to simplify the code a bit and to make the following changes easier.
- Implement method `write_current_buffer()` which writes content from the current time series buffer to a log file without running a thread in the background.  Use this to log data if the time span is short enough to be fully contained in the time series buffer.
- If the specified log file already exists, do not append to it anymore but overwrite the existing file.  The appending resulted in invalid files as the header was also written again and I anyway don't see a use-case where one would want to log independent sequences to the same file.
- Some bugfixes:
  - enable restarting after stopping
  - detect if stop is called without starting it first and don't attempt to write to the file in this case.

## How I Tested

With the `demo_data_logging.py` demo and some modifications of it to test the various cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,6 @@ set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/lib)
 set(CATKIN_PKGS ${CATKIN_PKGS}
   mpi_cmake_modules
   real_time_tools
-  mpi_cpp_tools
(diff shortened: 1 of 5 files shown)


luator:
@shrutij01 Do you have time to take a look at this?

shrutij01:
> @shrutij01 Do you have time to take a look at this?

Yes @luator I do I will take a look at this today

PR #76, opened 2020-08-17
Documentation

luator:

## Description

Add documentation.

Rendered version for review can be found here: https://people.tuebingen.mpg.de/felixwidmaier/robot_interfaces


docs/quick_start_example.md
```diff
@@ -2,16 +2,15 @@ Quick Start Example
 ===================
 
 Sending actions to and getting observations from the robot is very easy.  See
-the following example that simply sends a constant position command:
+the following example, using the TriFinger robot, that simply sends a constant
+position command:
```
mwuethri, line 6: i think it would be important to make a comment here that this is just an example of the frontend usage, and it will just keep waiting in this case, as no backend is started.

docs/timeseries.md
```diff
@@ -1,35 +1,65 @@
...
+hardware) goes through so called time series. When calling
+`append_desired_action(action)`, the action is not applied immediately but is
+*appended* to the time series of desired actions which serves as a queue.
+
+At each time step, identified by a _time index t_, the backend takes the action
+at position _t_ from the "desired actions" time series and sends it to the robot
+driver.  At the same time an observation is acquired from the robot and added to
+the "observation" time series. This means that the effect of the desired action
+`a_t` is not yet visible in the observation `y_t` as is illustrated below.
+(`a'_t` corresponds to the *applied action*, see [Desired vs Applied Action]())
```
mwuethri, line 25: this is displayed in a strange way, not sure if that's intended

docs/index.md
```diff
@@ -1,12 +1,32 @@
...
 
 Content
 -------
 
-- [Build Instructions](@ref md_docs_installation)
-- [Quick Start Example](@ref md_docs_quick_start_example)
-- [About Time Series](@ref md_docs_timeseries)
+- @ref md_docs_installation
+- @ref md_docs_quick_start_example
+- @ref md_docs_timeseries
```
mwuethri, line 22: i would put the timeseries before the quick start example, since this example cannot be understood just like that. i would also change the title of this, since it does not only describe the timeseries.

i would maybe call it Logic of Actions and Observations

docs/timeseries.md
```diff
@@ -63,6 +95,11 @@ auto t = frontend.append_desired_action(zero_torque_action);
 auto first_observation = frontend.get_observation(t);
 ```
 
+Note that the creation of the zero torque action in the above example is
+specific to the _TriFinger_ robot.  For other robots, the creation of the action
+would need to be adjusted to the action type of that specific robot.  Also, it m
+
```
mwuethri, line 101: sentence ends in the middle

docs/installation.md
```diff
@@ -14,12 +21,12 @@ separately:
```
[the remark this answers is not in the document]
luator: Good question.  I wasn't aware that catkin made it into the Ubuntu repos so that it could actually be installed without ROS.  I'm not sure if we have any actual dependency on ROS packages.

luator:
Addressed all comments and added missing doxygen in RobotFrontend.

PR #81, opened 2020-09-08
Store robot log in binary format

luator:
## Description

Add option to store the robot log in binary format using serialization by cereal.

This results in significantly smaller log files (~100 MB instead of ~240 MB for 2 minutes on TriFingerPro) and is easier to load in C++ applications.

This is is added in addition to the CSV-log (so all existing functionality is preserved) and only available in the "write current buffer" variant.

## How I Tested

By producing both csv and binary log on a test run on a TriFingerPro platform, converting the binary to csv and verifying that it results in the exact same values as when directly logging to CSV.  The script for converting will be added in a coming PR to robot_fingers.

### DIFF
--- a/include/robot_interfaces/robot_log_entry.hpp
+++ b/include/robot_interfaces/robot_log_entry.hpp
@@ -1,0 +1,39 @@
+/**
+ * \file
+ * \copyright Copyright (c) 2020, Max Planck Gesellschaft.
+ */
+#pragma once
+
+#include <time_series/interface.hpp>
+
+#include <robot_interfaces/status.hpp>
+
+namespace robot_interfaces
+{
+/**
+ * @brief Robot log entry used for binary log format.
+ *
+ * Contains all the robot data of one time step.
+ */
+template <typename Action, typename Observation>
+struct RobotLogEntry
+{
+    time_series::Index timeindex;
+    time_series::Timestamp timestamp;
+    Status status;
+    Observation observation;
+    Action desired_action;
+    Action applied_action;
+
+    template <class Archive>
+    void serialize(Archive& archive)
+    {
+        archive(timeindex,
+                timestamp,
+                status,
+                observation,
+                desired_action,
+                applied_action);
+    }
+};
+}  // namespace robot_interfaces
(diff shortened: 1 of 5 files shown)


include/robot_interfaces/robot_log_reader.hpp
```diff
@@ -0,0 +1,65 @@
...
+namespace robot_interfaces
+{
+/**
+ * @brief Read the data from a robot log file.
+ *
+ * The data is read from the specified file and stored to the `data` member
+ * where it can be accessed.
+ */
+template <typename Action, typename Observation>
+class RobotLogReader
```
shrutij01, line 27: I think we should specify this reader is for binary file type o/w it may be a bit confusing

include/robot_interfaces/robot_log_entry.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace robot_interfaces
+{
+/**
+ * @brief Robot log entry used for binary log format.
+ *
+ * Contains all the robot data of one time step.
+ */
+template <typename Action, typename Observation>
+struct RobotLogEntry
```
shrutij01, line 19: same for here if this is also specific for binary datatype

shrutij01 (commented):
Just a couple of minor comments wrt naming since it may not be obvious that the data will be logged in binary format only so it may be good to be explicit about it. Everything else looks great to me. ;)

include/robot_interfaces/robot_log_reader.hpp
luator: done

include/robot_interfaces/robot_log_entry.hpp
luator: This is currently only used for the binary log but the struct itself is not specific to that, it could also be used for other formats (e.g. with cereal it would be easy to use JSON or XML instead of the binary format).  Therefore, I would not add the "Binary" here.

shrutij01: Okay, yes, that makes sense

shrutij01 (approved):
looks great :)

PR #85, opened 2020-11-02
RobotLogger: Detect if robot data is empty.

luator:
## Description

Detect if the robot data is empty (i.e. robot did not start) in the
`write_current_buffer*` methods as otherwise it will hang there forever.

## How I Tested

With submission system setup on TriFingerPro (only explicitly tested `write_current_buffer_binary`).


shrutij01 (approved):
lgtm! :)

PR #89, opened 2020-11-27
Warning hunt in unit-tests + ROS2 packaging

MaximilienNaveau:

## Description

[//]: # "Description of what you did.  If it is more complex, consider to add a"
[//]: # "'Summary' section."

ROS2 packaging: using ament and colcon and a little bit of warning hunt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,96 +1,141 @@
-######################
-# set up the project #
-######################
-cmake_minimum_required(VERSION 3.1)
+#
+# Copyright (c) 2019, New York University and Max Planck Gesellschaft.
+#
+# License BSD-3 clause
+#
+
+#
+# Set up the project.
+#
+cmake_minimum_required(VERSION 3.10.2)
 
 project(robot_interfaces)
 
-# required to use std::shared_ptr in code and to link the python bindings
-if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,96 +1,141 @@
...
 
-##########################
-# building documentation #
-##########################
-add_doxygen_documentation()
+#
+# python bindings.
+#
+
+macro(ADD_PYTHON_BINDINGS module_name)
```
luator, line 94: @MaximilienNaveau What is the reason for not using `pybind11_add_module` anymore?
While porting other packages, I was using the following (based on what we had for catkin, I only needed to update the install destination) which works for me and is much shorter:

```cmake
macro(add_pybind11_module module_name)
    pybind11_add_module(${module_name} srcpy/${module_name}.cpp)
    target_include_directories(${module_name} PUBLIC include)
    target_link_libraries(${module_name} PRIVATE ${ARGN})
    install(TARGETS ${module_name}
        DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}")
endmacro()
```

The install destination below should be changed to `"${PYTHON_INSTALL_DIR}/${PROJECT_NAME}"` for the targets to end up at the right place (right now imports like `import robot_interfaces.py_generic` are not working anymore).

I can fix this as I'm anyway working on migrating my other packages.  I'm wondering, though, if there is some problem with the shorter macro, which I am not aware of.

Another question:  As we don't have pybind11_catkin anymore (which is where the above macro was defined), where should I put this now?  Would it fit into mpi_cmake_modules or should I rather create a new package for it?

PR #91, opened 2021-01-18
Make Status::error_message fixed-size

luator:

## Description

For use in a multi-process time series, the `Status` objects need to
have fixed size.  Thus replace the variable-size `std::string` with a
fixed size `char` array.  The length is set to 64 characters which is
more than enough for our current error messages (typically < 40 chars).

Since working with plain char-arrays is pretty annoying, make it private
and only access it through methods that convert the message from/to
`std::string`.

## How I Tested

Tested on the robot in a setup similar to the submission system.
The problem could be seen in situations where there is an error reported by the backend.  Without this fix, the frontend crashes when reading the corresponding status object as due to the suddenly added error message it exceeds the expected size.  With this fix, the status object with the error message can be read correctly.

### DIFF
--- a/include/robot_interfaces/robot_backend.hpp
+++ b/include/robot_interfaces/robot_backend.hpp
@@ -287,7 +287,7 @@ private:
 
                 robot_data_->status->append(status);
 
-                std::cerr << "Error: " << status.error_message
+                std::cerr << "Error: " << status.get_error_message()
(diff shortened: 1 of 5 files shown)


luator:
@vincentberenz Did you already have time to look at this?

vincentberenz (approved):
lgtm

PR #94, opened 2021-02-02
RobotFrontend: Remove deprecated methods

luator:
## Description

Remove some methods from `RobotFrontend` which are deprecated since a while.
For all of them there is an equivalent alternative with only the underscore removed in `time_stamp`/`time_index`, so in case they were still be used somewhere, fixing the code that used it is trivial.

## How I Tested
Compiled the workspace and searched the whole workspace to make sure the removed methods are not mentioned anywhere.


luator:
@vincentberenz Are you using the `RobotFrontend` somewhere?  If yes, this might be relevant for you.

PR #100, opened 2021-04-06
Use internal buffer in RobotLogger

luator:

## Description

Refactor the `RobotLogger` and add a logging mode that uses an internal buffer (same as in SensorLogger).  This should resolve some issues we had in the past where some time steps were missing in the log because they where not in the robot data time series anymore.

This new logging method supports writing to either binary, CSV or gzipped CSV files.

The old `start()`/`stop()` methods (used for continuous writing to file) have been renamed to `start|stop_continuous_writing()`.  Further this methods are marked deprecated as I plan to remove them soon, to not make the logger too bloated (this logging method is anyway not used at the moment).

To avoid confusion with the internal buffer, rename the `write_current_buffer*` methods to `save_current_robot_data*`.  The old names are kept as aliases for now but are marked deprecated.

### Other Changes
- Use applied_action in RobotFrontend::wait_for_timeindex.  The observation is the first time series to be filled by the backend, so when waiting only for this, it is not guaranteed that the other time series have already been filled.  To ensure this, use the applied_action time series instead, which is the last one to be filled.
- Avoid blocking in SensorLogger in case the backend has stopped.

## How I Tested

By running the unit tests.

### DIFF
--- a/tests/test_robot_logger.cpp
+++ b/tests/test_robot_logger.cpp
@@ -134,7 +134,7 @@ protected:
     }
 };
 
-TEST_F(TestRobotLogger, write_current_buffer)
+TEST_F(TestRobotLogger, save_current_robot_data)
 {
     Logger logger(data, 0);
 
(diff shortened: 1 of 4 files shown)


luator:
Rebased and adjusted to changes on master.

tests/test_robot_logger.cpp
```diff
@@ -54,21 +60,84 @@ class TestRobotLogger : public ::testing::Test
...
+
+        // check data
+        for (uint32_t i = start_t; i < end_t; i++)
+        {
+            int time_index;
+            double time_stamp, action_repetitions, error_status, obs_0, obs_1,
+                applied_0, applied_1, desired_0, desired_1;
+
+            ASSERT_TRUE(std::getline(infile, line))
+                << "Failed to read line " << i;
```
vincentberenz, line 122: I am bit confused: in which stream this is writing ?

vincentberenz (approved):
lgtm

tests/test_robot_logger.cpp
luator: This is the error message that is displayed if the assert fails.

PR #101, opened 2021-04-07
Refactor Gzip Stream Wrappers

luator:

## Description

- Move the code for wrapping streams with gzip (de-)compression to a separate library to avoid code duplication.
- For the istream identify if the input is actually gzip compressed by checking the first two bytes for the magic number.  This allows reading both compressed and uncompressed logs and thus provides backward compatibility.

## How I Tested

By reading both compressed and uncompressed logfiles and by running the unit tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,9 +12,6 @@ cmake_minimum_required(VERSION 3.10.2)
 project(robot_interfaces)
 
 # Using C++17.
-if(NOT CMAKE_C_STANDARD)
-  set(CMAKE_C_STANDARD 99)
-endif()
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
 endif()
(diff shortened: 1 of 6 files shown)


luator:
@vincentberenz @MaximilienNaveau Do you think the gzip_iostream library could be useful in other packages as well?  If yes, it would probably make sense to move it to a core package (not sure if there already exists one where it would fit in...).

vincentberenz:
> @vincentberenz @MaximilienNaveau Do you think the gzip_iostream library could be useful in other packages as well? If yes, it would probably make sense to move it to a core package (not sure if there already exists one where it would fit in...).

Definitely. Something I proposed earlier, but I never had the time to get to it, is moving the serialization from shared memory to an independent package. This could really be useful there.

luator:
We already have https://github.com/MPI-IS/serialization_utils, so if you think that makes sense, I can move it there.

vincentberenz:
> We already have https://github.com/MPI-IS/serialization_utils, so if you think that makes sense, I can move it there.

+1

MaximilienNaveau:
Sounds good!

PR #105, opened 2021-04-22
fix: Do not block in RobotLogger::stop() if buffer is empty

luator:
## Description
Do not wait for `oldest_timeindex()` when starting the logger.  Instead, simply assume that the first step is `t = 0` if there is no data yet.

This resolves an issue that `stop()` was blocking forever when the logger was stopped before any data was received.

Add unit tests to cover this case.

## How I Tested

By running the unit tests and by using it on the submission system with a user code that does not send any actions (and thus never starts the robot).


vincentberenz (approved):
lgtm

PR #108, opened 2021-05-14
fix: Do not block sensor logger if there is no data

luator:
## Description

Instead of blocking and waiting for the oldest time index, simply start
with `t = 0` if there is no data yet in the beginning.  The same fix has
already been applied recently on the RobotLogger class and prevents the
`stop()` method from blocking forever in case the logger is stopped
before any data is received.

## How I Tested

By running robot_fingers/trifinger_data_backend.py (without any other scripts, so no data is added to the time series) and stopping it.  Without this fix it was blocking in the "save camera log" step, with the fix it doesn't.

### DIFF
--- a/include/robot_interfaces/sensors/sensor_logger.hpp
+++ b/include/robot_interfaces/sensors/sensor_logger.hpp
@@ -145,7 +145,13 @@ private:
     //! Get observations from sensor_data_ and add them to the buffer.
     void loop()
     {
-        auto t = sensor_data_->observation->oldest_timeindex();
+        // Get the oldest available timeindex as starting point of the log.  In
+        // case there is no data yet (t == EMPTY), start with t = 0.
+        auto t = sensor_data_->observation->oldest_timeindex(false);
+        if (t == time_series::EMPTY)
+        {
+            t = 0;
+        }
 
         while (enabled_)
         {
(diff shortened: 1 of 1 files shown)


include/robot_interfaces/sensors/sensor_logger.hpp
```diff
@@ -145,7 +145,13 @@ class SensorLogger
     //! Get observations from sensor_data_ and add them to the buffer.
     void loop()
     {
-        auto t = sensor_data_->observation->oldest_timeindex();
+        // Get the oldest available timeindex as starting point of the log.  In
+        // case there is no data yet (t == EMPTY), start with t = 0.
+        auto t = sensor_data_->observation->oldest_timeindex(false);
```
vincentberenz, line 150: why using 'auto' for t ? Why not time_series::Index ?

vincentberenz (approved):
lgtm, just had one minor comment

include/robot_interfaces/sensors/sensor_logger.hpp
luator: I think this still dates back to the time series from real_time_tools, where the index type was a bit more complex.  I agree that it's better to be explicit here.

PR #138, opened 2026-03-11
Use real-time thread in SensorBackend

luator:
## Description

Even if the sensor itself may not be fully real-time capable (e.g. due to USB connection), it might still help to reduce timing variations.

## How I Tested

By recording tricamera log files and analysing the timing in them.
I observed that with the real-time thread, there are a few delay spikes that are larger as without but overall the timing is much more constant than with a normal thread:

The plots show the difference between camera timestamps of consecutive observations.

With regular threads (old behaviour):
old_2

With real-time thread (this PR):
new_rt_2

I'm not sure why the max spikes are larger in the RT case but since they are few, I think it's better than the overall inconsistent timing of the old implementation.  Anyway, I'll further investigate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@
   joints in place).
 - If an error message given to `Status::set_error()` is cut due to being too
   long, this is now indicated by setting '~' as last character.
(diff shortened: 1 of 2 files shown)


luator:
Ignoring broken CI build.

